PR #14, opened 2023-11-06
add a python3 jupyter kernel to shiny modules qmd

chendaniely:
The rendering process was picking up a different jupyter kernel for me by defualt.
All of the `.qmd` files should probably have a `jupyter: python3` line in it, but fixing this one specifically got me to build the site properly.

### DIFF
--- a/docs/workflow-modules.qmd
+++ b/docs/workflow-modules.qmd
@@ -1,5 +1,6 @@
 ---
 title: Shiny Modules
+jupyter: python3
 ---
 ## Introduction
 


schloerke:
Superseded by #439, which applies this same fix as a single top-level `jupyter: python3` key in `_quarto.yml` — covering all 10 pages that contain executable Python cells rather than one file.

For the record on why this one can't be merged as-is:

- It targets `docs/workflow-modules.qmd`, which no longer exists; the modules docs are now `docs/modules.qmd` and `docs/module-communication.qmd`.
- The branch is `CONFLICTING` against `main`.
- The diff inserts `jupyter: python3` above the opening `---`, so it would render as literal body text instead of parsing as front matter.

Thanks for the original report — the underlying diagnosis was right, and #439 carries it forward. Closing in favor of that.

PR #15, opened 2023-11-08
Add input components

skaltman:


components/inputs/selectize-multiple.qmd
```diff
@@ -29,133 +28,66 @@ listing:
...
-                    return render.DataTable( 
-                        penguins, 
-                        width = "500px", #<<
-                        height = "200px" #<<
-                    )
-
-            app = App(app_ui, server)
-
+        
+        <p>To access the value of a selectize list:</p>
```
garrettgman, line 82:
```suggestion
        <p>The currently selected value of the selectize list is accessible as a reactive value within the server function. To access the value of a selectize list:</p>
```
This may not be the best way to word it, but I think we should mention that you can only access the value from within the server function.

components/inputs/selectize-multiple.qmd
```diff
@@ -29,133 +28,66 @@ listing:
...
+        
+        <ol>
+          <li>
+          Use `input.<selectize_id>()` (e.g., `input.selectize()`) to access the selected value(s). The server value of a selectize list is a list of strings.
+          </li>
+        </ol>
+        
+        <p>See also: Selectize (Single) (TODO: ADD LINK TO Selectize (Single) HERE)</p>
```
garrettgman, line 90:
```suggestion
        <p>See also: Selectize (Single) (TODO: ADD LINK TO Selectize (Single) HERE) and Select (Multiple) (TO DO ADD SELECT MULTIPLE LINK HERE). The Select Input is similar to the Selectize Input, but _________.</p>
```
I don't know how to explain the difference between Select and Selectize, but the details page is likely where people will look for the explanation.

components/inputs/selectize-multiple.qmd
```diff
@@ -29,133 +28,66 @@ listing:
...
 ---
```
garrettgman, line 91: If I understand @gregswinehart correctly, we should add 
```
      variations:
        - title: null
          description: null
          preview: null
          code: null
```
To the end of pages that have no variations.

garrettgman: For this page though, I think we should add a variation that shows a Selectize that groups the items to be selected into groups with titles, as you mentioned in the details section. It is unique enough to count as a variation, and different enough that seeing an example would be useful for comprehension. To be clear, I wouldn't remove the description from the details section. I'd just supplement it with a variation.

garrettgman (approved):
@skaltman this looks really good. I left a few (hopefully helpful) suggestions.

gregswinehart (approved):

PR #16, opened 2023-11-08
Component: output plot matplotlib

chendaniely:


components/outputs/plot-matplotlib.qmd
```diff
@@ -0,0 +1,93 @@
...
+        - title: matplotlib.pyplot.subplots
+          href: https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.subplots.html
+          signature: matplotlib.pyplot.subplots(nrows=1, ncols=1, *, sharex=False, sharey=False, squeeze=True, width_ratios=None, height_ratios=None, subplot_kw=None, gridspec_kw=None, **fig_kw)
+        - title: matplotlib.axes.Axes.hist
+          href: https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.hist.html
+          signature: Axes.hist(x, bins=None, range=None, density=False, weights=None, cumulative=False, bottom=None, histtype='bar', align='mid', orientation='vertical', rwidth=None, log=False, color=None, label=None, stacked=False, *, data=None, **kwargs)
+
+      details: |
+
+              <p>The Matplotlib is a library that can be used to create plots.</p>
```
skaltman, line 59:
```suggestion
              <p>Matplotlib is a library that can be used to create plots.</p>
```

components/outputs/plot-matplotlib.qmd
```diff
@@ -0,0 +1,93 @@
...
+                <ul>
+                  <li>`@render.plot()`</li>
+                </ul>
+
+                <li>Pass `ui.output_plot()` the name of the function that builds the figure for the `input` argument,
+                as a character string,
+                to indicate which figure should be displayed in the div.
+                </li>
+              </ol>
+
```
skaltman, line 89: It might be useful to include a See also section, and then link to the other plot pages. I've been doing it like [this](https://github.com/posit-dev/py-shiny-site/pull/15/files#diff-c04103c288e8de2e806e3bff96a5971f36552997f5d52670b678456ec4a12767R84).(Greg said he can add links in later).

components/outputs/plot-matplotlib.qmd
```diff
@@ -0,0 +1,93 @@
...
+              <p>Follow three steps to display a Matplotlib figure in your app:</p>
+
+              <ol>
+                <li>
+                Add `ui.output_plot()` to the UI of your app to create a div in which to display the figure.
+                Where you call this function will determine where the figure will appear within the layout of the app.
+                </li>
+                <li>
+                Define a function within the `server()` function that creates the figure and
+                is wrapped around with the `@render.plot()` decorator to return the figure.
```
skaltman, line 70: I was a little confused about how to use `@render.plot()` here because it is mentioned in two places. I think it makes more sense to take out this mention of `@render.plot()` but keep the one below.

components/outputs/plot-matplotlib.qmd
```diff
@@ -0,0 +1,93 @@
...
+                    <li>The name of the decorated function (or `@output(id=...)`) should match the `id` of a `output_plot()` container.</li>
+                  </ul>
+
+                To notify Shiny that your function returns a reactive figure to display, decorate the function with:
+
+                <ul>
+                  <li>`@render.plot()`</li>
+                </ul>
+
+                <li>Pass `ui.output_plot()` the name of the function that builds the figure for the `input` argument,
```
skaltman, line 84: I found it a little confusing that the instructions return to `ui.output_plot()` at the end. Maybe you could say to pick an id in the first step ("Add `ui.output_plot()` to the UI"), and then mention that the function needs to have the same name. But what you have now is still clear.

skaltman (commented):
This looks good! I added some comments

components/outputs/plot-matplotlib.qmd
skaltman: I now see that the second mention of `@render.plot()` is a sub-bullet of this one, but I still find it a little confusing.

PR #25, opened 2023-11-17
Avoid Stamen basemaps in ipyleaflet example

cpsievert:
This example is currently broken (I'm pretty sure because a recent ipyleaflet release removed the Stamen basemaps)


gadenbuie:
Some background: Stamen basemaps are now hosted by Stadia, which requires an API key or an account with Stadia. I collected details about the transition in this issue in [rstudio/leaflet](https://github.com/rstudio/leaflet/issues/883#issuecomment-1749394106).

gadenbuie:
Some background: Stamen basemaps are now hosted by Stadia, which requires an API key or an account with Stadia. I collected details about the transition in this issue in [rstudio/leaflet](https://github.com/rstudio/leaflet/issues/883#issuecomment-1749394106).

PR #29, opened 2023-11-20
Add learn-shiny article on custom components.

nstrayer:
This is a fairly straightforward copying of the readme for the repo https://github.com/nstrayer/pyshiny-output-binding-example.

Adds a new article in the "In Depth" section of "Learn Shiny" that covers building a one-off javascript component for an app. Future work will add more info on packaging these custom components into a proper python package.


docs/custom-components.qmd
```diff
@@ -0,0 +1,314 @@
...
+
+# The problem
+
+You found a new table library that you really want to use in your Shiny app. Problem is, there's no wrapper for it, currently. The library is [Tabulator](https://tabulator.info/) and it's a javascript library for making tables with data.
+
+# The solution
+
+To implement a custom tabulator element for your app, you'll need to write three things:
+
+1. A javascript script that renders the element on the client side using the Tabulator library
```
gshotwell, line 19: Is it javascript or JavaScript?

docs/custom-components.qmd
```diff
@@ -0,0 +1,314 @@
...
+tabulator/
+  tabulator_esm.min.js
+  tabulator.min.css
+  tableComponent.js
+```
+
+Both `tabulator_esm.min.js` and `tabulator.min.css` are downloaded from [tabulator's website.](https://tabulator.info/docs/5.5/install#sources-download) `tableComponent.js` is the script that we will write that contains the code for rendering the table to our Shiny app.
+
+:::{.callout-note}
+The code in this article will be abbreviated to show the relevant parts. If you want to see the full code, see the [accompanying repo.](https://github.com/nstrayer/pyshiny-output-binding-example)
```
gshotwell, line 40: I think we probably want to move this to posit-dev if it's going to be cannonical

docs/custom-components.qmd
```diff
@@ -0,0 +1,314 @@
...
+}
+
+// Register the binding
+Shiny.outputBindings.register(
+  new TabulatorOutputBinding(),
+  "shiny-tabulator-output"
+);
+```
+
+This class has two methods that we need to implement: `find()` and `renderValue()`. The `find()` method is used to find the element that we want to render the table in. The `renderValue()` method is used to render the table in the element. After making that class we need to register it with Shiny so it can find and send data to instances of our output.
```
gshotwell, line 62:
```suggestion
This class has two methods that we need to implement: `find()` and `renderValue()`. The `find()` method is used to identify the element that will contain the rendered table. The `renderValue()` method is used to render the table in the element. After making that class we need to register it with Shiny so it can find and send data to instances of our output.
```

docs/custom-components.qmd
```diff
@@ -0,0 +1,314 @@
...
+
+### The `find()` method
+
+Now that we have the scaffolding setup, we can fill it in. Starting with the `find` method, this function is passed a `scope` object, which is a `JQuery` selection and should return the element you wish to render your output into.
```
gshotwell, line 66: This might have a bit too much passive voice, maybe something like :

"Now that we have the scaffolding setup we can start by filling in the find method. This function takes a `scope` object, which is a `JQuery` selection that should rereturn the element which will contain the output". I'm not sure if "contains the output" is the same as "render into".

docs/custom-components.qmd
```diff
@@ -0,0 +1,314 @@
...
+        return scope.find(".shiny-tabulator-output");
+    }
+
+    renderValue(el, payload) {...}
+}
+
+Shiny.outputBindings.register(...);
+```
+
+Note that we're using the class `".shiny-tabulator-output"` here to mark the element that we want to render the table in. This is the same class that we will use in our `output_tabulator()` function in our app's server code. You can use any valid css selector here, but it's common to use a descriptive class name like this.
```
gshotwell, line 80:
```suggestion
Note that we're using the class `".shiny-tabulator-output"` here to mark the element that we want to render the table in. This is the same class that we will use in our `output_tabulator()` function in our app's server code. You can use any valid css selector here, but it's common to use a class name which describes the output.
```

docs/custom-components.qmd
```diff
@@ -0,0 +1,314 @@
...
+
+### The `renderValue()` method
+
+Next, we fill in the main logic for rendering our table in to the `renderValue` method. This method gets passed two arguments: `el` which is an HTMLElement as found by our find function, and `payload` which is the data that the server has provided from the render function (more on this [soon.](#the-render_tabulator-decorator))
```
gshotwell, line 84:
```suggestion
Next, we fill in the main logic for rendering our table in to the `renderValue` method. This method gets passed two arguments: `el` which is an HTMLElement identifies by our find function, and `payload` which is the data that the server has provided from the render function (more on this [soon.](#the-render_tabulator-decorator))
```

docs/custom-components.qmd
```diff
@@ -0,0 +1,314 @@
...
+      });
+    }
+}
+
+Shiny.outputBindings.register(...);
+```
+
+The implementation of this function is not terribly important and draws directly from the [tabulator docs](https://tabulator.info/docs/5.5/quickstart).
+
+What matters is that we take our data, transform it in some way, and then instantiate our table with the `new Tabulator(el, {...})` call. In this case we take data in the form of the rows of a passed data frame, the column names, and the types of those columns (we decide this format when we [create the render decorator](#the-render_tabulator-decorator)), and construct a js object in the form of `data = [{col1: foo1, col2: bar1, ...}, {col1: foo2, col2: bar2, ...}]`. We also combine the column names and types to create the `columnsDef` object that Tabulator expects.
```
gshotwell, line 131:
```suggestion
What matters is that we take our data, transform it in some way, and then instantiate our table with the `new Tabulator(el, {...})` call. In this case we take data in the form of the rows of a passed data frame, the column names, and the types of those columns (this is determined by the [the render decorator](#the-render_tabulator-decorator)), and construct a js object in the form of `data = [{col1: foo1, col2: bar1, ...}, {col1: foo2, col2: bar2, ...}]`. We also combine the column names and types to create the `columnsDef` object that Tabulator expects.
```

docs/custom-components.qmd
```diff
@@ -0,0 +1,314 @@
...
+
+The format of this will vary entirely based upon the type of component you're building though, so if you don't follow, don't worry!
```
gshotwell, line 133:
```suggestion
Don't worry too much about following this particular format because it will change depending on the component that you are wrapping.
```

docs/custom-components.qmd
```diff
@@ -0,0 +1,314 @@
...
+
+:::{.callout-note}
+
+Since this code is relying on the `Shiny` object just existing in the Javascript context. It's safe to wrap all the above code in an if statement so it only runs if that object exists. This is useful if you're writing a package that might be used in a non-Shiny context, your code won't error out and break the document.
```
gshotwell, line 137:
```suggestion
Since this code is relying on the `Shiny` object existing in the Javascript context. It's safe to wrap all the above code in an if statement so it only runs if that object exists. This is useful if you're writing a package that might be used in a non-Shiny context because your code won't error out and break the document.
```

docs/custom-components.qmd
```diff
@@ -0,0 +1,314 @@
...
+    Shiny.outputBindings.register(...);
+}
+```
+:::
+
+To see the full javascript code, see `tabulator/tableComponent.js` in the [accompanying repo.](https://github.com/nstrayer/pyshiny-output-binding-example/blob/main/tabulator/tableComponent.js)
+
+## The `output_tabulator()` function
+
+For the case of our table we just need an HTML element to target with our javascript code. When we setup [the find method](#the-find-method) for our binding, we chose the class `shiny-tabulator-output` as the mark of a tabualtor output, so we need to add that class. We also need to allow the user to set the ID of the element so that Shiny knows which element to target with which output. By wrapping the `id` argument in `resolve_id()` we make sure it will work in the context of modules. We'll also add a height argument so that the user can set the height of the table.
```
gshotwell, line 152:
```suggestion
Next we need an HTML element to target with our javascript code. When we setup [the find method](#the-find-method) for our binding, we chose the class `shiny-tabulator-output` as the mark of a tabualtor output, so we need to add that class. We also need to allow the user to set the ID of the element so that Shiny knows which element to target with which output. By wrapping the `id` argument in `resolve_id()` we make sure it will work in the context of modules. We'll also add a height argument so that the user can set the height of the table.
```

docs/custom-components.qmd
```diff
@@ -0,0 +1,314 @@
...
+    return ui.div(
+        tabulator_dep,
+        # Use resolve_id so that our component will work in a module
+        id=resolve_id(id),
+        class_="shiny-tabulator-output",
+        style=f"height: {height}",
+    )
+```
+
+We use the `HTMLDependency` function to bind up the assets needed for tabulator that we made in the previous step and making sure that they're included in our app anytime the `output_tabulator()` function is called (but not more than once.)
```
gshotwell, line 179: I think we can include this in the callout

docs/custom-components.qmd
```diff
@@ -0,0 +1,314 @@
...
+        "columns": res.columns.tolist(),
+        "type_hints": res.dtypes.astype(str).tolist(),
+    }
+```
+
+The `output_transformer` decorator is a decorator factory that takes a function that returns a dictionary of data to be passed to the client side. The function that it decorates is passed two arguments: `_meta` and `_fn`.
+
+`_meta` is a dictionary of metadata about the function that is being decorated. We don't use it in our example.
+
+`_fn` is the function that is being decorated. Aka the function that goes below the `@render_tabulator()` in your app's server code. In this case we are expecting that that function returns either a pandas dataframe or `None`.
```
gshotwell, line 229:
```suggestion
`_fn` is the decorated function which appears below the `@render_tabulator()` decorator in your app's server code. In this case we are expecting the function to return either a pandas dataframe or `None`.
```

docs/custom-components.qmd
```diff
@@ -0,0 +1,314 @@
...
+
+In the code above we use types so that we can get some type checking in our IDE, but these are not required. Also note that the decorated function is an async function, so we need to use the `await` keyword when we call it for `resolve_value_fn()`.
+
+```python
+...
+res = await resolve_value_fn(_fn)
+...
+```
+
+`resolve_value_fn()` is a helper provided in `shiny.render.transformer` for resolving the value of a function that may or may not be async. This allows us to write our code in a way that is agnostic to how the user has written their render function.
```
gshotwell, line 239:
```suggestion
`resolve_value_fn()` is a helper provided in `shiny.render.transformer` for resolving the value of a function that may or may not be async. This ensures that the function will work whether or not the end user has defined their render function asynchronously.
```

docs/custom-components.qmd
```diff
@@ -0,0 +1,314 @@
...
+
+Next, we check to make sure that the value returned by the function is a dataframe. If it's not, we throw an error. This is not required, but it's good practice to do so.
```
gshotwell, line 241:
```suggestion
Next, we check to make sure that the value returned by the function is a dataframe. If it's not, we throw an error. This is not strictly required, but is a best practice.
```

docs/custom-components.qmd
```diff
@@ -0,0 +1,314 @@
...
+You found a new table library that you really want to use in your Shiny app. Problem is, there's no wrapper for it, currently. The library is [Tabulator](https://tabulator.info/) and it's a javascript library for making tables with data.
+
+# The solution
+
+To implement a custom tabulator element for your app, you'll need to write three things:
+
+1. A javascript script that renders the element on the client side using the Tabulator library
+2. An `output_tabulator()` function for placing the element in your app's UI
+3. A `render_tabulator()` decorator for passing table data to the javascript code rendering the element on the server side
+
```
gshotwell, line 22: I think it would be nice here to have a mermaid diagram of how all the components work together and what they do.

gshotwell (changes requested):
This is great overall, a couple of writing style things, but overall I think it looks good.

docs/custom-components.qmd
nstrayer: I've seen it both but switched it to "JavaScript" for all uses.

nstrayer:
I ended up bringing this article over in #46 to simplify things. Still using comments from review there though.

PR #36, opened 2023-11-29
Fix typos in venv installation instructions in install.qmd

ergodiclife:
Fix minor typos in venv installation instructions.


wch:
Hi, thanks for the PR. The use of `venv` instead of `.venv` was not a typo -- it was intentional. That said, it does look like the Python community does somewhat favor using `.venv`, so I think it's reasonable to take this change.

ergodiclife:
Hi Winston, thanks for your comment. It was mainly so that the code was consistent with the existing comment, which mentions `.venv`.

PR #42, opened 2023-12-04
Streamlit comp updates

gshotwell:
This updates the streamlit-comparisson article to use shiny express. 

TODO:
- [] Update to use Garrick's code comparison pattern


docs/apps/comp-streamlit/slider-update/classic.py
```diff
@@ -0,0 +1,81 @@
...
+        return dist_plot(filtered_data())
+
+    @output
+    @render.plot
+    def scatter():
+        return scatter_plot(filtered_data(), input.smoother())
+
+
+app = App(app_ui, server)
+from pathlib import Path
```
cpsievert, line 40: Looks like this example got repeated twice?

cpsievert:
Any idea why there doesn't seem to be a deployment preview available?

docs/helpers.py
```diff
@@ -0,0 +1,106 @@
...
+    # Additional files need to start with ## file:
+    for x, y in zip(path_list, file_names):
+        block.append_file(x, y)
+
+    # Finish with the closing tag
+    block.append("```")
+    return block
+
+
+def express_tabs(path: str) -> None:
```
cpsievert, line 88: Seems like passing the `height` along could be useful?

docs/helpers.py
```diff
@@ -0,0 +1,106 @@
...
+    exclusions: List[str] = [],
+    components: str = "editor, viewer",
+    viewer_height: str = "800",
+    extra_object: Any = "",
+) -> None:
+    print(
+        _include_shiny_folder(
+            path, file_name, exclusions, components, viewer_height, extra_object
+        )
+    )
```
cpsievert, line 40: Feels like this should get removed (and if you want to `print()` for local debugging, just `print(block)` at the end of the function)?

docs/apps/comp-streamlit/penguins/classic.py
```diff
@@ -0,0 +1,36 @@
...
+        ui.input_slider("mass", "Mass", 2000, 8000, 6000),
+        ui.input_checkbox("smoother", "Add Smoother"),
+    ),
+    ui.card(ui.output_plot(id="scatter")),
+    ui.card(ui.output_plot(id="mass_distribution")),
+)
+
+
+def server(input, output, session):
+    infile = Path(__file__).parent / "penguins.csv"
```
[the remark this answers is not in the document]
gshotwell: I'm actually not sure why it uses the file, the package is much better I'll switch it

docs/helpers.py
gshotwell: Oh this is designed to include a single shinylive app in a block. So you might want to use an app in tabs, but maybe also refer to just one version of the file in a standalone app without tabs. I added some doc strings to clarify that.

gshotwell:
I think we need to build the dev version of shinylive, which Winston was going to work on. We definitely should have that for the whole transition branch so we can look for bugs.

cpsievert:
@gshotwell I just realized the [`docs/helpers.py`](https://github.com/posit-dev/py-shiny-site/pull/42/files#diff-957a06be80c8a9090c95f54413f81719564dbb5c9a8e360f39d311f87bd58593) here has some overlap with what @gadenbuie is proposing in #66. 

Maybe it would make sense to use the infrastructure in #66 here as well? I think this article will also need some updates to reflect recent changes in Express, like `from shiny.express import layout` -> `from shiny.express import ui`

cpsievert:
@gshotwell just a heads up, I made some fixes and updates in [e3bca01](https://github.com/posit-dev/py-shiny-site/pull/42/commits/e3bca018e87810b72ebe2f00a82ab0bf68706e65).

gshotwell:
@cpsievert your changes look good, I pushed a couple of fixes to avoid duplicate "Since," sentences. I think the best thing to do is merge #66, then I'll refactor this to use that pattern.

docs/helpers.py
```diff
@@ -0,0 +1,137 @@
...
+    This function is designed to be called within a python chunk whose output is set
+    to be "asis" with `#| output: asis`.
+
+    Args:
+        path (str): The path to the Shiny application files.
+        viewer_height (str): The height of the resulting Shinylive viewer
+    """
+    block = QuartoPrint(
+        [
+            "::: {.panel-tabset group='express-switcher'}",
```
gadenbuie, line 116:
```suggestion
            "::: {.panel-tabset .shiny-mode-tabset group='shiny-app-mode'}",
```

or maybe put the `.shiny-mode-tabset` behind a flag that's enabled by default?

docs/helpers.py
```diff
@@ -0,0 +1,137 @@
...
+        extra_object (Any, optional): Any extra object to include. Defaults to "".
+    """
+    print(
+        _include_shiny_folder(
+            path, file_name, exclusions, components, viewer_height, extra_object
+        )
+    )
+
+
+def express_tabs(path: str, viewer_height: str = "400px") -> None:
```
gadenbuie, line 101: I recommend making this name more specific, since it creates Express/Core tabs containing complete shinylive editors.

gadenbuie (commented):
`helpers.py` looks great! I made a few suggestions to take advantage of the patterns in #66.

Also worth considering that we can use `py-shinylive` to create shinylive URLs or to manage the shinylive bundles as needed. There are semi-internal helpers that we could build on (as the need arises). See https://github.com/posit-dev/py-shinylive/pull/20

docs/helpers.py
gshotwell: This seems good to me, we can always refactor if we want the flag.

docs/helpers.py
gshotwell: Good call

PR #46, opened 2023-12-06
Add documentation on building custom components for apps and distributing as packages

nstrayer:


nstrayer:
It's not _critical_ but some of the stuff in the article would benefit by https://github.com/posit-dev/py-shiny/pull/874 being merged so that the templates the user's would get when "playing along" match the article code more closely.

docs/custom-component-one-off.qmd
```diff
@@ -0,0 +1,323 @@
...
+        });
+        return obj;
+      }
+
+      // Render the table
+      new Tabulator(el, {
+        data: data.map(zipRowWithColumns),
+        layout: "fitColumns",
+        columns: columnsDef,
+      });
```
wch, line 131: I might be missing something, but how does simply instantiating the `Tabulator` object render it into the page?

If I am missing something, it might help to explain how that happens.

docs/custom-component-one-off.qmd
```diff
@@ -0,0 +1,323 @@
+---
+title: "Custom javascript component"
```
wch:
```suggestion
title: "Custom JavaScript component"
```

docs/custom-component-one-off.qmd
```diff
@@ -0,0 +1,323 @@
...
+        return scope.find(".shiny-tabulator-output");
+    }
+
+    renderValue(el, payload) {...}
+}
+
+Shiny.outputBindings.register(...);
+```
+
+Note that we're using the class `".shiny-tabulator-output"` here to mark the element that we want to render the table in. This is the same class that we will use in our `output_tabulator()` function in our app's server code. You can use any valid css selector here, but it's common to use a descriptive class name like this.
```
wch, line 89:
```suggestion
Note that we're using the class `".shiny-tabulator-output"` here to mark the element that we want to render the table in. This is the same class that we will use in our `output_tabulator()` function in our app's server code. You can use any valid CSS selector here, but it's common to use a descriptive class name like this.
```

docs/custom-component-one-off.qmd
```diff
@@ -0,0 +1,323 @@
...
+
+### The `renderValue()` method
+
+Next, we fill in the main logic for rendering our table in to the `renderValue` method. This method gets passed two arguments: `el` which is an HTMLElement as found by our find function, and `payload` which is the data that the server has provided from the render function (more on this [soon.](#the-render_tabulator-decorator))
```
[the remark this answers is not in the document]
nstrayer: ✅

docs/custom-component-one-off.qmd
```diff
@@ -0,0 +1,323 @@
...
+}
+
+Shiny.outputBindings.register(...);
+```
+
+Note that we're using the class `".shiny-tabulator-output"` here to mark the element that we want to render the table in. This is the same class that we will use in our `output_tabulator()` function in our app's server code. You can use any valid css selector here, but it's common to use a descriptive class name like this.
+
+### The `renderValue()` method
+
+Next, we fill in the main logic for rendering our table in to the `renderValue` method. This method gets passed two arguments: `el` which is an [HTMLElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement) as found by our find function, and `payload` which is the data that the server has provided from the render function (more on this [soon](#the-render_tabulator-decorator)).
```
wch, line 93:
```suggestion
Next, we fill in the main logic for rendering our table in to the `renderValue` method. This method gets passed two arguments: `el`, which is an [HTMLElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement) as found by our find function, and `payload`, which is the data that the server has provided from the render function (more on this [soon](#the-render_tabulator-decorator)).
```

docs/custom-component-one-off.qmd
nstrayer: Added a comment. You _can_ hold onto the object and make updates but we dont do that here. Just instantiating like this will render the table, though.

docs/custom-component-one-off.qmd
```diff
@@ -0,0 +1,323 @@
...
+
+What matters is that we take our data, transform it in some way, and then instantiate our table with the `new Tabulator(el, {...})` call. In this case we take data in the form of the rows of a passed data frame, the column names, and the types of those columns (we decide this format when we [create the render decorator](#the-render_tabulator-decorator)), and construct a js object in the form of `data = [{col1: foo1, col2: bar1, ...}, {col1: foo2, col2: bar2, ...}]`. We also combine the column names and types to create the `columnsDef` object that Tabulator expects.
+
+The format of this will vary entirely based upon the type of component you're building though, so if you don't follow, don't worry!
+
+:::{.callout-note}
+
+Since this code is relying on the `Shiny` object just existing in the Javascript context. It's safe to wrap all the above code in an if statement so it only runs if that object exists. This is useful if you're writing a package that might be used in a non-Shiny context, your code won't error out and break the document.
```
wch, line 148:
```suggestion
This code relies on the `Shiny` object just existing in the JavaScript context, but you may want to wrap all the above code in an `if (Shiny)` statement so it only runs if the `Shiny` object exists. This is useful if you're writing a component that might be used in a non-Shiny context, your code won't error out and break the document.
```

docs/custom-component-one-off.qmd
```diff
@@ -0,0 +1,325 @@
...
+    Shiny.outputBindings.register(...);
+}
+```
+:::
+
+To see the full JavaScript code, see `tabulator/tableComponent.js` in the [accompanying repo.](https://github.com/posit-dev/pyshiny-output-binding-example/blob/main/tabulator/tableComponent.js)
+
+## The `output_tabulator()` function
+
+For the case of our table we just need an HTML element to target with our JavaScript code. When we setup [the find method](#the-find-method) for our binding, we chose the class `shiny-tabulator-output` as the mark of a tabualtor output, so we need to add that class. We also need to allow the user to set the ID of the element so that Shiny knows which element to target with which output. By wrapping the `id` argument in `resolve_id()` we make sure it will work in the context of modules. We'll also add a height argument so that the user can set the height of the table.
```
wch, line 163:
```suggestion
For the case of our table, we just need an HTML element to target with our JavaScript code. When we set up [the find method](#the-find-method) for our binding, we chose the class `shiny-tabulator-output` as the mark of a Tabulator output, so we need to add that class. We also need to allow the user to set the ID of the element so that Shiny knows which element to target with which output. By wrapping the `id` argument in `resolve_id()` we make sure it will work in the context of Shiny modules. We'll also add a `height` parameter so that the user can set the height of the table.
```

Should also link to the Shiny modules page here.

docs/custom-component-one-off.qmd
```diff
@@ -0,0 +1,325 @@
...
+    return ui.div(
+        tabulator_dep,
+        # Use resolve_id so that our component will work in a module
+        id=resolve_id(id),
+        class_="shiny-tabulator-output",
+        style=f"height: {height}",
+    )
+```
+
+We use the `HTMLDependency` function to bind up the assets needed for tabulator that we made in the previous step and making sure that they're included in our app anytime the `output_tabulator()` function is called (but not more than once.)
```
wch, line 190:
```suggestion
We use the `HTMLDependency` function to bind up the assets needed for tabulator that we made in the previous step and to make sure that they're included in our app any time the `output_tabulator()` function is called (but not more than once).
```

docs/custom-component-one-off.qmd
```diff
@@ -0,0 +1,325 @@
...
+
+:::{.callout-note}
+Note the use of `all_files=True` here. This makes it so we can do the esm import of the tabulator library. Otherwise `tabulator_esm.min.js` would not be hosted and the js library couldn't find it.
+:::
+
+Now, the `output_tabulator()` function can be called anywhere we want to render a table in our app.
+
+## The `render_tabulator()` decorator
+
+Now we've got the client-side logic finished, we need to write a custom render decorator that sends our data into the component.
```
wch, line 200: I think this paragraph should add a sentence about what a render function does: it takes the value returned by the user's function, and transforms it into an object which is then converted to JSON and then sent to the client, where it is received by the output binding's `renderValue` method. (And it's OK if the "transformation" is just returning the object unchanged.)

gshotwell:
I tried building a JS wrapper and I have a few thoughts which are probably a bit bigger than this PR. So I pulled them into #48, I think this PR is a big improvement so we shouldn't block it with a big rewrite, but I think we do probably want to break this education task into smaller pieces down the road.

docs/custom-component-one-off.qmd
```diff
@@ -0,0 +1,325 @@
...
+        id=resolve_id(id),
+        class_="shiny-tabulator-output",
+        style=f"height: {height}",
+    )
+```
+
+We use the `HTMLDependency` function to bind up the assets needed for tabulator that we made in the previous step and making sure that they're included in our app anytime the `output_tabulator()` function is called (but not more than once.)
+
+:::{.callout-note}
+Note the use of `all_files=True` here. This makes it so we can do the esm import of the tabulator library. Otherwise `tabulator_esm.min.js` would not be hosted and the js library couldn't find it.
```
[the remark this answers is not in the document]
nstrayer: It probably should be.

docs/custom-component-one-off.qmd
```diff
@@ -0,0 +1,325 @@
...
+    }
+```
+
+The `output_transformer` decorator is a decorator factory that takes a function that returns a dictionary of data to be passed to the client side. The function that it decorates is passed two arguments: `_meta` and `_fn`.
+
+`_meta` is a dictionary of metadata about the function that is being decorated. We don't use it in our example.
+
+`_fn` is the function that is being decorated. Aka the function that goes below the `@render_tabulator()` in your app's server code. In this case we are expecting that that function returns either a pandas dataframe or `None`.
+
+In the code above we use types so that we can get some type checking in our IDE, but these are not required. Also note that the decorated function is an async function, so we need to use the `await` keyword when we call it for `resolve_value_fn()`.
```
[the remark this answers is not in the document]
nstrayer: I moved some stuff around and added a callout about the async-ness. It's a bit awkward because it's required to allow the user to be uninhibited with their function return type, but in this article, we are the user so in theory that flexibility is not needed.

gshotwell (commented):
I thought I submitted these earlier, so feel free to ignore them if the draft has moved past them. The main suggestion is to use the templating flag `shiny create -t js-react` to get people right to the template.

docs/custom-components-pkg.qmd
```diff
@@ -0,0 +1,405 @@
...
+
+While there are lots of ways to develop components with live-feedback (e.g. Storybook, dev servers like `vite`, etc) an easy way to develop a component with our package structure is to use the example app, an editable mode pip install, and the watch mode for our build step. We can do this with the following steps:
+
+1. Install the package in [“editable mode”](https://setuptools.pypa.io/en/latest/userguide/development_mode.html) with `pip install -e .`
+
+2. Run the bundler in watch mode with `npm run watch`. This will watch the `srcts` directory for changes and automatically rebuild the javascript when it detects a change.
+
+3. Run the example app in live-reload mode. If you're using VScode, the [Shiny for Python extension](https://marketplace.visualstudio.com/items?itemName=Posit.shiny-python) enables this automatically when pressing the run button above the app script.
+
+Now you can update your component javascript/python functions and your app will automatically reload with the changes. Happy developing!
```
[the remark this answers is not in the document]
nstrayer: I would but Pypi has frozen new account development since I started this article so I don't actually have the ability to test it out. I'll put a link about how to do it, though.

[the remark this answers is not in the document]
nstrayer: Added link out to the Python Packaging User Guide which honestly is most likely a much better resource than I could have hoped to write given how terse it already is.

docs/custom-components-pkg.qmd
```diff
@@ -0,0 +1,405 @@
...
+      onNewValue={(color) => onNewValue(color)}
+    />
+  ),
+});
+
+// Color Picker React component
+function ColorPickerReact(...) { ... }
+```
+
+In here we declare the `tagName` of for our component. This name is used to generate the correct markup with python later. Under the hood `makeReactInput()` generates a [webcomponent](https://developer.mozilla.org/en-US/docs/Web/Web_Components) to hold our input. A webcomponent is a custom HTML element. The `tagName` provided here represents the name of that custom component we're generating. In this case we're registering the component as `<fancy-color-picker>`.
```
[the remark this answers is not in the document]
nstrayer: Add this but reworded it a bit to reflect how webcomponents work a bit better

jcheng5 (approved):

PR #47, opened 2023-12-09
Layout component blog post

chendaniely:
files are definitely in the wrong place, but this is for review


gshotwell:
I think this should probably go in https://github.com/rstudio/shiny-dev-center/?

chendaniely:
closing. submited the post to the correct repo

PR #51, opened 2023-12-19
Change code blocks to not run code

wch:
I noticed that when building the docs, a couple of pages used Jupyter to run R and Python code, even though it wasn't actually necessary to run code. 

This PR changes it so that the it doesn't have to run any code, but keeps the highlighting. This makes the build time a bit faster.

### DIFF
--- a/docs/workflow-modules.qmd
+++ b/docs/workflow-modules.qmd
@@ -52,8 +52,7 @@ You can use this function in combination with list comprehension to further redu
 
 A simple function cleans up your code, but still requires multiple calls to that function.
 
-```{python}
-#| source-line-numbers: "9-13"
+```{.python}
 from shiny import App, render, ui
 
 def my_slider(id):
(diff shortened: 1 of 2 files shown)


docs/workflow-modules.qmd
```diff
@@ -52,8 +52,7 @@ You can use this function in combination with list comprehension to further redu
 
 A simple function cleans up your code, but still requires multiple calls to that function.
 
-```{python}
-#| source-line-numbers: "9-13"
```
wch, line 56: I noticed that this highlighting wasn't actually used because the `line-highlight` extension isn't enabled for this page, and when I added the extension, it highlighted what was probably the wrong set of lines.

But if line highlighting is desirable here, we could change change things to enable the extension and fix the highlighted lines.

This is what it currently looks like on the web site:

image

https://shiny.posit.co/py/docs/workflow-modules.html

gshotwell (approved):

PR #53, opened 2023-12-19
Re-organizing the component gallery supporting files

gadenbuie:
### Display messages

- [x] display-messages/modal.qmd
- [x] display-messages/notifications.qmd
- [x] display-messages/progress-bar.qmd
- [x] display-messages/tooltips.qmd

### Outputs

- [x] outputs/data-grid.qmd
- [x] outputs/datatable.qmd
- [x] outputs/image.qmd
- [x] outputs/map-ipyleaflet.qmd
- [x] outputs/plot-matplotlib.qmd
- [x] outputs/plot-plotly.qmd
- [x] outputs/plot-seaborn.qmd
- [x] outputs/text.qmd
- [x] outputs/ui.qmd
- [x] outputs/value-box.qmd
- [x] outputs/verbatim-text.qmd

### Inputs

- [x] inputs/action-button.qmd
- [x] inputs/action-link.qmd
- [x] inputs/checkbox-group.qmd
- [x] inputs/checkbox.qmd
- [x] inputs/date-range-selector.qmd
- [x] inputs/date-selector.qmd
- [x] inputs/numeric-input.qmd
- [x] inputs/password-field.qmd
- [x] inputs/radio-buttons.qmd
- [x] inputs/select-multiple.qmd
- [x] inputs/select-single.qmd
- [x] inputs/selectize-multiple.qmd
- [x] inputs/selectize-single.qmd
- [x] inputs/slider-range.qmd
- [x] inputs/slider.qmd
- [x] inputs/switch.qmd
- [x] inputs/text-area.qmd
- [x] inputs/text-box.qmd

PR #55, opened 2024-01-02
Use quarto v1.4.533

gadenbuie:
Updates the site to use the v1.4.533 preview release of Quarto.

This should have a small impact on the site and is worth it for a few reasons:

1. quarto 1.4 uses Bootstrap 5.3, which is the same version used by py-shiny and bslib.
2. Bootstrap 5.3 brings sweeping improvements and we should be building on these as much as possible; delaying the update increases the chances of introducing avoidable technical debt in the process. Improvements include wider use of CSS variables and dark mode styles.
3. There are a number of bugs that will be fixed by 1.4 that we'll be stuck with in 1.3; we're also pushing the envelope of Quarto's capabilities and may discover new bugs in the process (that would only be fixed with 1.4).
	* The biggest bug we have with 1.3 is that EJS templates eat empty new lines, causing the code examples in the components gallery to be smushed together. This is fixed in 1.4.


gshotwell:
I'm not against this, but for posterity what do we need from the dev version?

gadenbuie:
@gshotwell I updated the OP to describe what we get from this. Also #53 is dependent on this PR.

@gshotwell @gregswinehart I've looked through a selection of pages and fixed anything I noticed, most of which I found in the components gallery. Let me know if you see anything and I'll be happy to tweak or address.

gshotwell (approved):

PR #64, opened 2024-01-12
Express Release: Docs and site changes

gadenbuie:
Feature branch to collect all Express-related documentation and website updates.

For posit-dev/py-shiny#969

## Dev Notes

* Target any Express-related updates on this branch.
* Please prefer leaving commit history in tact and using merge commits. Alternatively use PR branch for git history revisions.

## Previewing

The shinylive examples in the deployment previews use the current PyPi release of packages, but the examples and writing target the new release, so many of them will be broken. If you'd like to locally build the site with the latest versions of packages, do:

```bash
# First, make sure you can build shinylive
git clone https://github.com/posit-dev/shinylive
cd shinylive
make submodules
make all

# Link shinylive CLI to freshly built assets
pip install shinylive
shinylive assets link-from-local ./build

# Re-build with latest dev versions
cd packages
cd py-htmltools && git pull && cd ..
cd py-shiny && git pull && cd ..
cd py-shinywidgets && git pull && cd ..
cd py-faicons && git pull && cd ..
cd ..
make clean && make all

# Finally, build py-shiny-site from this branch
git clone https://github.com/posit-dev/py-shiny-site
cd py-shiny-site
git checkout rc-express
make serve
```


cpsievert:
@gadenbuie @gregswinehart any idea why something [like this](https://github.com/posit-dev/py-shiny-site/pull/64/files#diff-2f50e6a1e6c047fee5fc091c6b9a2b7b12494af87c16b3cec07b0cd53d2de983R18) is leading to those CSS classes being added to _both_ the `` and the figure div? When I build locally, those classes are only added to the ``?

gregswinehart:
@cpsievert  @gadenbuie  Looks like something I added while working on the homepage carousel. I'll fix it!

docs/overview.qmd
```diff
@@ -3,232 +3,560 @@ title: Overview
...
-ui.input_numeric("n", "N", 0)
+Once you've [installed](install-create-run.qmd) Shiny, the `shiny create` CLI command provides access to a collection of useful starter templates. This command walks you through a series of prompts to help you get started quickly with a helpful example. One great option is the dashboard template, which can be created with:
+
+```bash
+shiny create -t dashboard
 ```
 
-Adding a `ui.panel_sidebar` call around `ui.input_numeric` simply wraps the control with additional HTML (a `<div>` and `<form>`, in this case).
-The nesting of the generated HTML matches the nesting of the Python function calls in your UI.
+![The resulting dashboard generated by the dashboard template](assets/dashboard-template.png){class="img-shadow"}
```
gshotwell, line 536: I don't think this screenshot matches the template so we should change the photo.

docs/express-vs-core.qmd
```diff
@@ -0,0 +1,262 @@
...
+In Shiny Express, you'll instead see:
+
+```python
+from shiny import reactive
+from shiny.express import input, render, ui
+```
+
+Note that both import `ui` and `render`, but from different places.
+While Express's `shiny.express.ui` has almost all of the same UI functions as Core's `shiny.ui`, their function signatures often differ slightly, to reflect Express's different usage patterns.
+And the `render` functions---well actually, they are identical right now, but we're planning to add some Express-specific features to the `shiny.express.render` versions in the near future.
```
gshotwell, line 54: I would cut this line as it'll become out of date, and the user doesn't really care that they're implemented the same way. Many of the `express.ui` functions are also re-exports.

docs/express-vs-core.qmd
```diff
@@ -0,0 +1,262 @@
...
+Finally, notice that Express also imports `input` from `shiny.express`.
+This isn't needed in Core, because the `input` object is passed into the server function as an argument.
+Since Express has no server function, we made it an attribute of `shiny.express`.
+
+### Organization of UI and server code
+
+**Every Shiny Core app file has a UI section, and a server section;** the two are separate and distinct.
+
+The UI section consists of a single (potentially very long and deeply nested) expression, stored as a variable named `app_ui` by convention.
+The object this produces is actually straight up HTML, which is sent to the browser when it first loads the app.
```
gshotwell, line 67:
```suggestion
The object this produces is actually simple HTML, which is sent to the browser when it first loads the app.
```

docs/express-vs-core.qmd
```diff
@@ -0,0 +1,262 @@
...
+        return f"Hello, world!\nIt's currently {time()}."
+
+app = App(app_ui, server)
+```
+
+**In Shiny Express, there isn't this hard distinction** between UI and server.
+
+Instead, everything coexists as top-level code in the app file: input components, layout directives, outputs (including their rendering logic), and reactive functions.
+
+Also, you aren't forced to combine your UI into a single complex object.
```
gshotwell, line 104: I would cut this, as part of our pitch for Core is that the UI object is simpler to read and maintain, aslo we say you can combine variables into a UI object. IMO it's enough just say that everything coexists.

docs/express-vs-core.qmd
```diff
@@ -0,0 +1,262 @@
...
+    reactive.invalidate_later(1)
+    return datetime.now()
+
+@render.code
+def greeting():
+    return f"Hello, world!\nIt's currently {time()}."
+```
+
+Again, notice how `greeting` in this app does _not_ have a corresponding call to `output_code("greeting")`.
+This is because in Shiny Express, `render` functions automatically add an output to the page---no need to do it manually.
```
gshotwell, line 129:
```suggestion
This is because in Shiny Express, `render` functions implicitly add an output to the page as a side effect so you don't need the output function.
```

docs/express-vs-core.qmd
```diff
@@ -0,0 +1,262 @@
...
+##### Core advantages
+
+- Because the UI structure is kept separate from the server, it is easier to read, reorder, and restructure. This advantage grows as app UIs grow larger.
+- Explicit server function declaration gives us a natural place to put code that should only execute at startup (top level) versus for each session (server function body). In contrast, in Express, all of the code in the app file is executed for each session.
+
+##### Express advantages
+
+- It's nice for beginners not to have to learn about the difference between UI and server.
+- Avoids having to write code in two different places for a single output, and having to make the IDs match up.
+- No need to write nested function declarations (i.e. functions inside the server function), which can be surprising to Python programmers.
```
gshotwell, line 140:
```suggestion
- No need to write nested function declarations (i.e. functions inside the server function)
```

gshotwell: This reads as though we shipped a bad framework and are fixing it, which I don't think is what we mean

docs/express-vs-core.qmd
```diff
@@ -0,0 +1,262 @@
...
+
+##### Core advantages
+
+- Passing children as arguments is beautifully simple and robust from a programming perspective.
+- Containers and non-containers behave very similarly (in fact, non-containers act no different than containers with no children).
+- All of the Core UI components are simple value-returning functions, without side effects. This makes them easy to compose, reuse, refactor, inspect, and test.
+
+##### Express advantages
+
+- The `with` block syntax is slightly more readable, as it depends only on indentation and not on parentheses and commas.
```
gshotwell, line 235: I don't love more readable since we say Core is more readable in other contexts. Maybe "More familiar to many users" or "easier to write because you don't have to worry about commas"?

docs/express-or-core.qmd
```diff
@@ -0,0 +1,267 @@
...
+
+Shiny Express is designed to get you up and running as quickly as possible.
+It's also designed to let you author your app with a minimum of boilerplate.
+As a result, it really shines in the early stages of both the learning journey and the lifecycle of an app.
+
+Compared to Express, Core enforces a more structured approach that requires more initial effort to use.
+However, that investment has its own payoff.
+
+The bulk of the rest of this article will focus on the advantages of Shiny Core.
+This isn't because we think it's better, but because Shiny Express's advantages---being more approachable and more concise---are fairly self-evident, while Shiny Core's advantages are more subtle.
```
gshotwell, line 21:
```suggestion
This isn't because we think it's better overall, but because Shiny Express's advantages---being more approachable and more concise---are fairly self-evident, while Shiny Core's advantages are more subtle.
```

docs/express-or-core.qmd
```diff
@@ -0,0 +1,267 @@
...
+
+But for larger and longer-lived apps, Shiny Core's more opinionated approach becomes an advantage.
+It is much easier to add, remove, or relocate pieces of your UI when all of its code is in one place, with no server code to confuse things. Similarly, when you're trying to understand the relationship between a reactive calculation and some outputs, it's much easier to do so when you don't have intermingled UI code in the way.
+
+At the [bottom of this page](#appendix), you'll find a Shiny Core version of the [dashboard app from the Essentials section](user-interfaces.qmd#all-together-now){target="_blank"} of this guide.
+Compare their respective source code, and consider:
+
+* Which version makes the structure of the UI more obvious?
+* Which version would make you more confident in moving UI elements around?
+* Which version makes it easier to understand the reactive calcs and outputs?
```
gshotwell, line 38:
```suggestion
* Which version makes it easier to understand the reactive calcs and outputs?
* Imagine you were picking up an app that had been written a year ago by someone else, which version would you prefer?
```

docs/express-or-core.qmd
```diff
@@ -0,0 +1,267 @@
...
+
+## Feature set
+
+**Consider using Shiny Core if you need to use Shiny Modules or dynamic UI.**
+
+At this time, Shiny Core's functionality is a superset of Express, meaning that anything you can do in Express, you can also do in Core.
+The reverse is not true, however.
```
gshotwell, line 45:
```suggestion
The reverse is not, however, true.
```

docs/express-or-core.qmd
```diff
@@ -0,0 +1,267 @@
...
+bill_rng = (min(tips.total_bill), max(tips.total_bill))
+
+ICONS = {
+    "user": fa.icon_svg("user", "regular"),
+    "wallet": fa.icon_svg("wallet"),
+    "currency-dollar": fa.icon_svg("dollar-sign"),
+    "gear": fa.icon_svg("gear")
+}
+
+app_ui = ui.page_sidebar(
```
gshotwell, line 105: Do we want to break this UI object into variables or use a function to generate the UI? It might be nice to demonstrate that functionality.

gshotwell (commented):
This looks great, a few review comments which are optional

docs/express-vs-core.qmd
jcheng5: I planned to change this line when we change that functionality.

docs/express-vs-core.qmd
jcheng5: It is simpler to read and maintain, and harder to write. Putting it in a single object is a downside--I felt it viscerally yesterday when porting the dashboard app.

docs/express-vs-core.qmd
jcheng5: I don't read it that way, and I think taking off the second half makes it worse, not better. Your version makes it sound like nested function declarations are an inherent disadvantage, mine just says it's "surprising" to Python programmers.

docs/express-vs-core.qmd
jcheng5:
```suggestion
- The `with` block syntax is more forgiving to write, as you don't need to separate elements with commas. (Every Shiny Core app author has probably seen the "Perhaps you forgot a comma?" error message many times.)
```

docs/express-or-core.qmd
jcheng5:
```suggestion
* Which version makes it easier to understand the reactive calcs and outputs?
* Imagine you were picking up an app that had been written a year ago by someone else. Which version would you prefer?
```

docs/express-or-core.qmd
jcheng5:
```suggestion
However, the reverse is not true.
```

docs/express-or-core.qmd
jcheng5: I didn't want to have to introduce that here, and definitely didn't want to _not_ introduce it and then use it.

Incidentally, I personally prefer it this way, one big expression unless the depth gets out of hand or there's something specific you want to isolate conceptually.

gadenbuie:
Link Issues

```
[ 45/342] docs/user-interfaces.qmd
WARNING: Unable to resolve link target: docs/reactivity.qmd

[ 46/342] docs/express-vs-core.qmd
WARNING: Unable to resolve link target: layouts/sidebars.qmd
WARNING: Unable to resolve link target: layouts/panels-cards.qmd
WARNING: Unable to resolve link target: layouts/arrange.qmd
WARNING: Unable to resolve link target: components/inputs/text-box.qmd
WARNING: Unable to resolve link target: components/outputs/plot-matplotlib.qmd

[ 61/342] docs/ui-overview.qmd
WARNING: Unable to resolve link target: docs/ipywidgets.qmd
WARNING: Unable to resolve link target: docs/ipywidgets.qmd
WARNING: Unable to resolve link target: docs/ipywidgets.qmd

[ 67/342] docs/express-in-depth.qmd
WARNING: Unable to resolve link target: api/reactive.file_reader.qmd
WARNING: Unable to resolve link target: api/reactive.poll.qmd
```

jcheng5 (approved):

PR #68, opened 2024-01-15
Overhaul "Learn Shiny" content to be Express focused

cpsievert:
To preview locally (with the shinylive extension using the latest versions of packages):

```bash
# First, make sure you can build shinylive
git clone https://github.com/posit-dev/shinylive
cd shinylive
make submodules
make all

# Link shinylive CLI to freshly built assets
pip install shinylive
shinylive assets link-from-local ./build

# Re-build with latest dev versions
cd packages
cd py-htmltools && git pull && cd ..
cd py-shiny && git pull && cd ..
cd py-shinywidgets && git pull && cd ..
cd py-faicons && git pull && cd ..
cd ..
make clean && make all

# Finally, build py-shiny-site from this branch
git clone https://github.com/posit-dev/py-shiny-site
cd py-shiny-site
git checkout learn-shiny-overhaul
make serve
```


_quarto.yml
```diff
@@ -12,7 +12,6 @@ project:
     - outputs
     - display-messages
     - docs
-    - in-depth
```
cpsievert, line 15: Removed `in-depth/` folder since the content seemed redundant and not used/finished

_quarto.yml
```diff
@@ -81,8 +78,6 @@ website:
         target: _blank
       - text: "Function Reference"
         href: api/index.qmd
-      - text: "Help"
-        href: docs/help.qmd
```
cpsievert, line 85: Both `help` and `install` got folded into the "Workflow" section

docs/assets/head-content.html
```diff
@@ -0,0 +1,30 @@
...
+      link.innerHTML = `<i class="bi bi-lightning-fill"></i> Run on shinylive`;
+      pre.appendChild(link);
+    });
+    // Add nav-underline to panel-underline where appropriate
+    document.querySelectorAll(".panel-underline .nav").forEach((x) => {
+      x.classList.remove("nav-tabs");
+      x.classList.add("nav-underline");
+      x.classList.add("justify-content-center");
+    });
+  });
```
cpsievert, line 29: TODO: this should probably get updated/removed in favor of infrastructure in #66

docs/jupyter-widgets.qmd
```diff
@@ -0,0 +1,263 @@
...
+```bash
+pip install shinywidgets
+pip install plotly ipyleaflet
+```
+
+## Get started {#get-started}
+
+Rendering ipywidgets with `{shinywidgets}` works like most other Shiny [outputs](quick-start.qmd#outputs).
+The main difference is that, instead of using a `@render.*` decorator, you'll use a `@render_widget` decorator.
+Let's consider an example of displaying a [plotly](https://plotly.com/python/plotly-express/) graph that reacts to changes in Shiny input controls:
```
cpsievert, line 33: TODO: expand on this a bit to incorporate different packages like altair, etc.

docs/jupyter-widgets.qmd
```diff
@@ -0,0 +1,263 @@
...
+from shiny import render
+import shiny.express
+
+@render_widget
+def map():
+    return ipyl.Map(zoom=2)
+
+@render.text
+def center():
+    cntr = reactive_read(map.widget, 'center')
```
cpsievert, line 178: TODO: Finish https://github.com/posit-dev/py-shinywidgets/pull/119 and update/proof-read writing

docs/install-create-run.qmd
```diff
@@ -149,37 +129,72 @@ You can also modify these settings on a per-file basis with comments at the top
...
+::: callout-tip
+### Copy/paste examples
 
-If your project contains a single Shiny app, you can replace the `"${file}"` in the above snippet with the relative path to your `app.py` file. With a hardcoded path like this, you can press `F5` to start debugging your app without having to activate the `app.py` editor tab first.
+If you find an example on this site that you want to run/edit locally, you can use `shiny create -t basic-app` to get a basic app template, and then copy/paste the code from the example into the template.
+:::
 
-Note that in debugging mode, edits to your source code will not result in automatic reloading (i.e. the `shiny run` command will not have `--reload`); instead, you will need to restart your debugging session (`Command+Shift+F5` / `Ctrl+Shift+F5`).
+
+## Run
```
gshotwell, line 155: This is all copied yeah? I believe you can include the contents of one document in the other so that there's only one place to update. Here's how I did that for a presentation: https://github.com/gshotwell/shiny-algorithm/blob/main/index.qmd#L276 but I think there's an easier way if you're not using reveal.js

docs/custom-components-pkg.qmd
```diff
@@ -282,7 +285,7 @@ By using the `resolve_id(id)` function here when declaring our ID, we make sure
 Like with the JavaScript, the process for setting up the python code for an output binding is not too different. Although there is a bit of extra work because we need to build both the ui _and_ server components. Here's how we would do that for the color shower output binding we defined above:
 
 ```{.python filename="fancy_color_picker.py"}
-@output_transformer()
+@output_transformer(default_ui=output_color)
```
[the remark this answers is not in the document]
cpsievert: I'm trying to keep changes here minimal for now since Barret is planning on updating this article with the latest `Renderer` stuff

docs/install-create-run.qmd
cpsievert: Oh sorry, you had reviewed `docs/create-run.qmd` which was supposed to be deleted. This is the actual file that gets included in the site (and I applied your other feedback to this file).

docs/reactive-patterns.qmd
```diff
@@ -0,0 +1,232 @@
...
+        # Read input.n() without taking a dependency on it
+        return f"Result: {input.n()}"
+```
+
+:::
+
+
+## Requiring input {#req}
+
+Sometimes you want to require an input before executing a reactive function and/or displaying output. The `req()` is designed for this purpose. It effectively stops execution until the statement provided is truthy.
```
gshotwell, line 95: Users will likely not know what truthy means.

docs/reactive-patterns.qmd
```diff
@@ -0,0 +1,232 @@
...
+
+@render.ui
+def clock():
+    return f"Current time: {cur_time()}"
+```
+
+
+## Reactive file reading {#file}
+
+Sometimes its useful invalidate a reactive function when a file changes.
```
gshotwell, line 155:
```suggestion
Sometimes its useful invalidate a reactive function when a file changes.
```
```suggestion
Shiny apps often depend on on-disk files and its useful invalidate a reactive function when those files change.
```

docs/reactive-patterns.qmd
```diff
@@ -0,0 +1,232 @@
...
+For example, lets extend the example from above to write the current time to a file every second, and then read and display the contents of that file:
```
gshotwell, line 156: This isn't that compelling of an example to me, it might be better to say something like "Imagine an app that depends on a csv file which is populated by an Airflow pipeline, you want the app to always show the most up-to-data data without needing to restart the user's session". the plot will update whenever the file changes on disk. You can't really show the interaction in the same way but I think it's better to have an example which solves a real user problem.

docs/reactive-patterns.qmd
```diff
@@ -0,0 +1,232 @@
...
+@render.ui
+@reactive.file_reader("time.txt")
+def time():
+    with open("time.txt") as f:
+        return f"Current time {f.read()}"
+```
+
+## Reactive polling {#poll}
+
+Sometimes it's useful to invalidate a reactive function on a schedule, but only _if a certain condition is met_.
```
gshotwell, line 191: We're starting too many sentences with `Sometimes it's useful` can we have a different introcution? You could maybe just start with the example

docs/reactive-patterns.qmd
```diff
@@ -0,0 +1,232 @@
...
+    reactive.invalidate_later(1)
+    return datetime.now().strftime('%H:%M:%S')
+
+@render.ui
+def clock():
+    return f"Current time: {cur_time()}"
+```
+
+
+## Reactive file reading {#file}
```
gshotwell, line 153: We have templates for three monitoring cases in case they're helpful https://github.com/posit-dev/py-shiny-templates

docs/reactive-patterns.qmd
```diff
@@ -0,0 +1,232 @@
...
+
+## Reactive polling {#poll}
+
+Sometimes it's useful to invalidate a reactive function on a schedule, but only _if a certain condition is met_.
+For example, suppose we want to check if a (potentially large) file (or database) has changed/updated every so often, and if it has, re-read it.
+The `@reactive.poll()` decorator is designed for this purpose.
+When applying the decorator, make sure to provide a function that is relatively cheap to execute, since it will be executed repeatedly on an interval.
+And, in the event that that function's value changes, the reactive function will be invalidated and re-executed.
+
+For example, lets extend the example from above to write the current time to a file every 0.5 seconds, but only read and display the contents every 2 seconds:
```
gshotwell, line 197: As above I think it's better to have a database query which generates a table and a plot, even if we have to tell instead of show what it does.

docs/reactive-patterns.qmd
```diff
@@ -0,0 +1,232 @@
...
+# Every 2 seconds, check if the file has changed.
+# If it has, re-read it, and display the contents.
+# Note: checking for the modified time of a file is cheap
+# compared to reading the file contents
+@render.ui
+@reactive.poll(lambda: os.path.getmtime("time2.txt"), 2)
+def time():
+    with open("time2.txt") as f:
+        return f"Current time {f.read()}"
+```
```
gshotwell, line 232: Probably need a concluding sentence here

gshotwell (commented):
Got up to the ui sections. I think this is great, but the reactive programming stuff still seems like it's not quite right for people learning about reactivity for the first time. I think we should ship them because they're still a big improvement, but we may want to revisit them again fairly soon.

docs/ui-customize.qmd
```diff
@@ -0,0 +1,256 @@
...
+#| viewerHeight: 150
+
+from shiny.express import ui
+
+with ui.card():
+    ui.card_header("Card header", class_="bg-primary-subtle lead")
+    "Card body"
+```
+
+### CSS Variables {#css-variables}
```
gshotwell, line 137: Would it be easy to generate a table of common variables?

docs/ui-dynamic.qmd
```diff
@@ -2,158 +2,111 @@
...
+Amongst these [dynamic UI](#dynamic-ui) is the most general and powerful, but also comes with the most overhead.
 
 
-## Updating inputs
+### Conditional UI
 
-One approach to dynamically changing inputs is to use `ui.update_select()`.
-
-Try clicking checkbox inputs in the example below, to add them to the dropdown.
+Often times it's useful to show or hide UI components based on some condition.
```
gshotwell, line 12:
```suggestion
The most basic way to create dynamic UIs is by conditionally hiding a UI element on the client side.
```

docs/ui-dynamic.qmd
```diff
@@ -2,158 +2,111 @@
...
+`ui.panel_conditional()` enables this by showing/hiding UI based on a whether a JavaScript condition is truthy.
```
gshotwell, line 13:
```suggestion
`ui.panel_conditional()` enables this by showing/hiding UI based on a JavaScript condition.
```

docs/ui-dynamic.qmd
```diff
@@ -2,158 +2,111 @@
...
+This condition can reference input values, and used to make any sort of UI conditional (not just a string as done here).
```
gshotwell, line 14:
```suggestion
This condition can reference input values, and can be used to make any sort of UI conditional.
```

docs/ui-dynamic.qmd
```diff
@@ -2,158 +2,111 @@
...
-
 
-## Dynamic UI from server output
+### Updating inputs
 
-The previous section showed how shiny can dynamically effect UI by updating an existing input.
-An alternative approach is to replace the UI input with `output_ui()`, and then generate the input on the server.
-
-This is shown below example---which uses `ui.output_ui()` and `@render.ui`.
+Updating inputs server-side is useful for updating input parameters that are usually hard-coded in the UI.
```
gshotwell, line 35:
```suggestion
In addition to hiding elements on the client side, you can also update elements from the server. This is used in cases where you want to change on part of a ui element without regenerating the whole element. For example you might want to change `ui.input_select` choices when a user takes a particular action.
```

docs/ui-dynamic.qmd
```diff
@@ -2,158 +2,111 @@
...
+    ui.update_selectize("x", choices=CHOICES[choices])
 ```
 
-Notice that the `ui.input_select()` was moved to inside a server output function (`def ui_select():`).
-Whenever `input.checkbox_item()` changes, the `ui.input_select()` is regenerated on the server and output.
 
+### Dynamic UI {#dynamic-ui}
 
-## Comparing approaches
+Dynamic UI is a powerful feature that enables dynamic (i.e., server-side) rendering of any UI component(s).
```
gshotwell, line 63:
```suggestion
Finally, `ui.output_ui` lets you generate a ui element entirely on the server side, which is an extremely flexible way to dynamically generate UI elements. 
```

docs/ui-dynamic.qmd
```diff
@@ -2,158 +2,111 @@
...
+Like any other output, start with a container for output (via `ui.output_ui()`), and then apply `@render.ui` to a function that returns UI component(s).
+Here, for a starting example, we'll recreate the previous example using dynamic UI.
```
gshotwell, line 65:
```suggestion
We can use this to recreate the previous example using dynamic UI.
```

docs/ui-dynamic.qmd
```diff
@@ -2,158 +2,111 @@
...
 
-Note two important points:
 
-* This string is a javascript expression.
-* Instead of using `input.radio()`--like in the server function--you use `input.radio` without parentheses to refer
-  to input values.
+::: callout-warning
+## Dynamic UI vs. updating inputs vs. conditional UI
 
+Dynamic UI is a more general mechanism than the updating inputs and conditional UI patterns, and can be used to update any UI component(s) (not just inputs). However, updating inputs is more efficient than dynamic UI, and should be preferred when possible.
```
gshotwell, line 111:
```suggestion
Dynamic UI is a more general mechanism than the updating inputs and conditional UI patterns, and can be used to update any UI component(s) (not just inputs). However, updating inputs is more efficient than dynamic UI, and should be preferred where possible.
```

docs/ui-html.qmd
```diff
@@ -1,117 +1,151 @@
 ---
-title: "UI and HTML"
+title: "UI as HTML"
+format: html
 ---
 
-The user interface (UI) of a Shiny application is HTML, but in most cases, you will use wrapper functions that make it easier to create the HTML. To get the most out of user interfaces, it's important to understand the relationship between Shiny's UI functions and the HTML that they generate.
 
-UI components are available via the `shiny.ui` submodule. At the most basic level, you can create HTML tags by calling a correspondingly named function. For example, to create a `<div>` tag, you would call `ui.div()`.
+Underneath the hood, Shiny UI stands on a foundation of HTML, CSS, and JavaScript.
```
gshotwell, line 7:
```suggestion
Under the hood, Shiny UI stands on a foundation of HTML, CSS, and JavaScript.
```

docs/ui-html.qmd
```diff
@@ -1,117 +1,151 @@
...
 ui.a("Help", href="help.html", class_="help-link")
+<a href="help.html">Help</a>
 ```
+:::
 
-For convenience, common tags like `div`, `span`, and `a` are available directly from the `ui` submodule. All HTML tags are also available via `ui.tags`, so if you need to use a less common tag like `li`, you can use `ui.tags.li()`
+## List fragments
 
-Sometimes it it useful to return several tag objects together, but in a way where they are not nested within another tag. You can use a `TagList()` for this:
+Sometimes it's useful to create a list fragment of HTML -- use a `TagList()` for this.
```
gshotwell, line 95: Probably need to explain why someone would do this

docs/ui-html.qmd
```diff
@@ -1,117 +1,151 @@
...
-#| components: [editor, cell]
-
-ui.input_checkbox("enable", "Enable?")
-```
-
-A complete UI for a Shiny app will put components in a `page_` function. Here is a simple page with a checkbox input and a text output.
-
-```{shinylive-python}
-#| components: [editor, cell]
+Sometimes it's useful to include content in the `<head>` of the document, like CSS, JavaScript, `<meta>` tags, etc.
```
gshotwell, line 109: We need an alternative to "Sometimes it's useful"

docs/ui-html.qmd
```diff
@@ -1,117 +1,151 @@
...
-ui.page_fluid(
-    ui.input_checkbox("enable", "Enable?"),
-    ui.output_text_verbatim("txt"),
+ui.head_content(
+    ui.tags.meta(name="robots", content="noindex")
 )
 ```
 
-Broadly speaking, there are three categories of UI components: layout components, inputs, and outputs. The `page_fluid()` is a layout component, and it contains one input and one output.
+Note that if you find yourself using `ui.head_content()` to import CSS and JavaScript that should only be included once, you might want to consider using `HTMLDependency()` instead (see below).
```
gshotwell, line 120: Probably should use `include_css()` or `include_js()`?

docs/ui-html.qmd
```diff
@@ -1,117 +1,151 @@
...
 
-:::{.callout-tip}
-## Shiny for Python compared to R
+## HTML Dependencies
 
-In Shiny for Python, the page-related functions start with `page_`, the input-related functions start with `input_`, and the output-related functions start with `output_`. This is different from Shiny for R, where the functions are named `fluidPage`, `checkboxInput`, and `textOutput`.
-:::
+Sometimes it's useful to have more reassurance that CSS/Javascript files are only loaded once, even they happen to be included in multiple places in the UI (e.g., loading a framework like React).
```
gshotwell, line 124: Alternative to Sometimes it's useful

docs/ui-layouts.qmd
```diff
@@ -1,7 +1,140 @@
...
+| name | description | filling |
+| ---- | ----------- | ------- |
+| `page_sidebar()` | A page with a sidebar. | ✔ |
+| `page_navbar()` | A multi-page application, with an optional sidebar. | ✔ |
+| `page_fillable()` | A basic fillable page. | ✔ |
+| `page_fluid()` | A "full width" page (with a small gutter on each side). | ✗ |
+| `page_fixed()` | A "restricted width" page (great for "articles"). | ✗ |
+| `page_bootstrap()` | A bare bones [Bootstrap](https://getbootstrap.com/docs/5.3/getting-started/introduction/) page. | ✗ |
+
+The first three page options are designed to fill the page (by default), meaning that if components like plots, cards, widgets, etc. appear as direct children of the page, they grow/shrink to fill the page.
```
gshotwell, line 20:
```suggestion
The first three page options are designed to fill the page (by default), meaning that if components like plots, cards, or widgets appear as direct children of the page, they will grow/shrink to fill the page.
```

docs/ui-layouts.qmd
```diff
@@ -1,7 +1,140 @@
...
+
+By default, the sidebar is hidden on mobile devices (and can be opened/closed via the caret icon).
+Here, we've set `open="always"` so both the sidebar and main content are always visible, which works great for small examples, but may not be ideal for apps with lots of content.
+:::
+
+
+
+### `page_navbar()`
+
+A multi-page application, with an optional sidebar.
```
gshotwell, line 75: Do we want to say more about this?

gshotwell (commented):
Looks great, one thing is that quite a lot of sentences start with some version of "it's often useful" or "it's sometimes useful". I think we can more or less get rid of all of these and just jump right into the content. I made some suggested changes, but flagged a few of them just to be revised. I think it's a good idea to use second person language in these cases "input.select is for when you want to select an input"

docs/quick-start.qmd
```diff
@@ -0,0 +1,488 @@
...
+
+## file: requirements.txt
+pandas
+```
+
+:::
+
+
+::: {.callout-tip}
+### Quarto integration
```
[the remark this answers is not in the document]
gshotwell: My vote is that we should remove it from the quick start until we actually get questions about it.

docs/ui-layouts.qmd
cpsievert: Ahh sorry, this was another file that should've been removed (we can lean on the layout gallery instead now).

docs/quick-start.qmd
```diff
@@ -0,0 +1,488 @@
...
+  - inputs.html
+  - outputs.html
+  - server.html
+---
+
+Welcome to the Shiny quick start guide!
+After reading this article, you'll be able to get started creating basic Shiny apps.
+In later articles, we'll build on these concepts to create more advanced apps, like the following dashboard:
+
+![A Shiny dashboard with visuals for exploring restaurant tips (covered in the [next article](dashboards.qmd)).](assets/tipping-dashboard.png){class="img-fluid shadow rounded"}
```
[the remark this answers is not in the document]
cpsievert: Isn't that kind of true of the first 4 sections of articles and why we're including the site-wide banner?

docs/quick-start.qmd
```diff
@@ -0,0 +1,488 @@
...
+Shiny also integrates well with [Quarto](https://quarto.org/), allowing you to leverage its web-based output formats (e.g., [dashboards](https://quarto.org/docs/dashboards/interactivity/shiny-python/index.html)) in combination with Shiny outputs and reactivity.
+:::
+
+
+### Reactivity {#reactivity}
+
+Shiny's reactive programming model is perhaps its most powerful and proven concept[^reactive-signals].
+Compared to Python frameworks like [streamlit](comp-streamlit.qmd) or dash, reactivity allows Shiny to automatically determine the minimal amount of work to keep things up-to-date, meaning you don't need to manually cache computations (which can be tedious and error-prone) to keep your apps performant.
+Also, by learning how reactivity powers basic apps, you'll find those same concepts extend naturally to more advanced apps.
+In short, Shiny knows to re-execute a _reactive functions_ (e.g., `render` functions) when any of its _reactive dependencies_ (i.e., `input`) change.
```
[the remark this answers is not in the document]
cpsievert: > I think if we're teaching them how it should be a longer piece with lots of graphs.

I was originally gonna go in this direction, but @jcheng5 gave me specific feedback not to do this. I did watch you video though and I think that content could work well in new a Overview/Motivation article at the start of the "Reactivity" section

docs/reactive-foundations.qmd
```diff
@@ -0,0 +1,151 @@
+---
+title: "Foundations"
```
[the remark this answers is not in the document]
cpsievert: I agree, a motivation article would be interesting https://github.com/posit-dev/py-shiny-site/pull/68#discussion_r1462455609

docs/reactive-foundations.qmd
```diff
@@ -0,0 +1,151 @@
...
+  - reactive-calculations.html
+  - reactive-events.html
+  - reactive-values.html
+---
+
+At the heart of Shiny is a reactive programming framework.
+We often refer to this framework as "good magic" because it's easy to get started with, but also decomposes into simple pieces which combine in powerful ways.
+
+In [Quick Start](quick-start.qmd), we covered the most common form of reactivity: changes in `input` causing relevant `render` functions to re-execute (aka **invalidate**).
+More generally, Shiny knows to re-execute **reactive functions** (e.g., `render` functions) when relevant **reactive dependencies** (e.g., `input`) change.
```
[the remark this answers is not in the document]
cpsievert: That's sort of the goal of this article: to loosely define and provide more examples of reactive functions and dependencies. I'd be happy to consider adding a more formal definition if you can think of one, but it's probably too late to avoid the terms altogether

docs/reactive-foundations.qmd
```diff
@@ -0,0 +1,151 @@
...
+title: "Foundations"
+format: html
+aliases:
+  - reactive-programming.html
+  - reactive-calculations.html
+  - reactive-events.html
+  - reactive-values.html
+---
+
+At the heart of Shiny is a reactive programming framework.
```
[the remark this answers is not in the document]
cpsievert: Can we leave this for a new Motivation/Overview article? I acknowledge and appreciate the desire to have motivating examples, but I also think is value in not obscuring the main idea with unnecessary context

docs/express-motivation.qmd
```diff
@@ -0,0 +1,343 @@
...
+  markdown:
+    wrap: sentence
+format:
+  html:
+    include-in-header: assets/head-content.html
+---
+
+So far, we've really only seen Shiny _Express_: a simple way to learn and create basic apps.
+As we'll discuss here, the same things that make Express simple also make it difficult to manage complexity when your app grows in size and sophistication.[^streamlit-comp]
+And, although we'll cover some techniques for managing complexity and adding sophistication to Express, when you find yourself reaching for these techniques, it might be time to switch to Shiny Core.
```
gshotwell, line 13:
```suggestion
In this article we'll cover some techniques for managing complexity and adding sophistication to Express, but when you find yourself reaching for these techniques too often it might be time to switch to Shiny Core.
```

docs/express-motivation.qmd
```diff
@@ -0,0 +1,343 @@
...
+Shiny Core is a slightly more involved way to write apps using the same [components](ui-components.qmd) and [reactivity](reactive-foundations.qmd) you've already learned.
```
gshotwell, line 14: I personally prefer the term "structured" to "involved" because I think it's a bit more accurate, but also not negative.

docs/express-motivation.qmd
```diff
@@ -0,0 +1,343 @@
...
+The main difference is that Shiny Core embraces concepts like [decoupling](#decoupling) and [functional programming](#functional) which enforces programming patterns that, although require more effort upfront, will make it easier to manage complexity as your app grows.
```
gshotwell, line 15:
```suggestion
The main difference between Shiny Core and Shiny Express is that Shiny Core embraces concepts like [decoupling](#decoupling) and [functional programming](#functional). These programming patterns require a little more effort upfront, will make it much easier to manage complexity as your app grows.
```

docs/express-motivation.qmd
```diff
@@ -0,0 +1,343 @@
...
+
+If you're already convinced that Shiny Core is right for your app, the [next article](express-difference.qmd) outlines the syntax differences and provides a guide for transitioning from Express to Core.
+
+[^streamlit-comp]: [Streamlit](https://streamlit.io/) is similar to Express in that it's easy to get started with, but difficult to manage complexity. See [this article](https://www.shinyapps.io/articles/streamlit-vs-shiny/) for a comparison between Streamlit and Shiny.
```
gshotwell, line 19: I think I might make this a callout pointing to another article. I don't think Streamlit users are going to be interested in Core at all, so by the time they get here they'll be on board with Shiny. Also Express way way better at managing complexity than Streamlit, so I probably wouldn't equate them. Express is a lot closer to Core on this metric than it is to Streamlit.

docs/express-motivation.qmd
```diff
@@ -0,0 +1,343 @@
...
+
+## Decoupling {#decoupling}
+
+A big reason for Express' simplicity is that things like render functions get displayed exactly where they appear in the code.
```
gshotwell, line 23:
```suggestion
Shiny Express apps are simple to write because render functions get displayed exactly where they appear in the code.
```

gshotwell: This might be too strong, but I think it's accurate. It seems like this was the main goal of Express and everything else kind of flowed from accomplishing that goal.

docs/express-motivation.qmd
```diff
@@ -0,0 +1,343 @@
...
+This is great for basic apps, since you don't have to separate backend from frontend (aka server from UI) logic.
```
gshotwell, line 24: This bugs me a bit because the reader probably doesn't really get front-end and back-end at this point. I think I might just assert that it's great for basic apps and not say why. So something like:

"This is great for simple apps, but can lead to code that's harder to reason about as your application grows in size and complexity."

docs/express-motivation.qmd
```diff
@@ -0,0 +1,343 @@
...
+However, this can lead to code that's harder to reason about when it grows in size and complexity.
+For example, consider this snippet Express app -- can you visualize, at a glance, how the overall UI is structured?
```
gshotwell, line 26: I like this question formulation a lot

docs/express-motivation.qmd
```diff
@@ -0,0 +1,343 @@
...
+            fig.legend(handles, labels, loc='upper center')
+
+            fig.tight_layout()
+            return fig
+        ui.input_select("species", "Species", ["Chinstrap", "Adelie", "Gentoo"])
+```
+</details>
+
+The `render.plot` function above has a lot going on, and so it's easy to overlook that there's a `ui.input_select()` which appears just below the plot, inside the same card.
+Also, with multiple levels of indentation, it's easy to mess up the identation; and as a result, have the input appear in the wrong place, or not work at all.
```
gshotwell, line 73:
```suggestion
Also, with multiple levels of indentation, it's easy to mess up the indentation and have the input appear in the wrong place, or not work at all.
```

docs/express-motivation.qmd
```diff
@@ -0,0 +1,343 @@
...
+
+Express does offer a way to workaround this problem: the `ui.hold()` context manager, which allows you to define a `render` function in one place, then display it in another.
+This, in a sense, is similar to what Shiny Core forces you to do: separate server from UI logic, and then connect them together with an `ui.output_*()` container element.
```
gshotwell, line 76: Does the reader know about `ui.output` at this point? Maybe we should add some text about this after line 15.

docs/express-motivation.qmd
```diff
@@ -0,0 +1,343 @@
...
+            fig.tight_layout()
+            return fig
+        ui.input_select("species", "Species", ["Chinstrap", "Adelie", "Gentoo"])
+```
+</details>
+
+The `render.plot` function above has a lot going on, and so it's easy to overlook that there's a `ui.input_select()` which appears just below the plot, inside the same card.
+Also, with multiple levels of indentation, it's easy to mess up the identation; and as a result, have the input appear in the wrong place, or not work at all.
+
+Express does offer a way to workaround this problem: the `ui.hold()` context manager, which allows you to define a `render` function in one place, then display it in another.
```
gshotwell, line 75: I think we need to include the Core syntax and why it's good here. Right now it reads:

1) Express has problem
2) Here's how to solve it in Express

I think the idea of this article is that it should motivate Shiny Core, so at the end the reader should know about Core syntax and want to give it a try. Could we do something like break the programming-with-express code into its own article and link to that?

gshotwell (changes requested):
I'm sorry to say I really don't like the direction that the Motivate express article has gone in. My main concern with Express is that we want to avoid abandoning Shiny Core, and so we need a very clear article which articulates why people should still care about and use Shiny Core. Currently the Express-Core translation article reads more like "Here's why you might think you need Shiny Core but really Express does all of those things too", and includes much more express code than Core code. My suggestion is:

- Write a new article called "programming with express" that covers `@expressify` `@hold` `@render.express` and talks about managing complexity in Express. 
- Reference that article in a callout, but have this article be all about the virtues of Core. 

If we don't have a clear articulation of why Core is good, Express looks a lot more like something that supersedes Core entirely than adjuvant syntax.

docs/express-introduction.qmd
```diff
@@ -0,0 +1,41 @@
...
+The next article illustrates how these downsides can be mitigated to some extent in Express with some advanced tooling.
+However, if you feel yourself reaching for these tools, it might be time to switch to Core.
+If you're already convinced that Shiny Core is right for your app, feel free to skip ahead to the [transitioning](express-to-core.qmd) article.
+
+::: callout-note
+### Shiny Core is a superset of Express
+
+Shiny Core is a superset of Express, meaning that anything you can do in Express, you can also do in Core.
+The reverse is not true, however.
+One example of this [Shiny Modules](modules.qmd), which are not currently supported in Express.
```
gshotwell, line 33:
```suggestion
For example [Shiny Modules](modules.qmd) are not currently supported in Express.
```

PR #70, opened 2024-01-18
Docs for ExtendedTask feature

jcheng5:
I feel like this might be a little too technical and confusing, maybe I should just focus on ExtendedTask? And have a minimal section at the end saying that async/await are supported in reactives but aren't useful for increasing responsiveness?

@cpsievert after this is merged to main, we'll also have to figure out where to park this article in the rc-express branch. It might make more sense in "Miscellaneous" since it's such a (relatively) advanced topic.


docs/reactive-nonblocking.qmd
```diff
@@ -0,0 +1,316 @@
...
+
+Worse, if you have multiple users, then one user's long-running operation will block the other users' apps from running as well.
+These other users will not even be aware that their slow performance is due to another user's actions.
+
+In this article, we'll look at how to make your Shiny apps more responsive by using non-blocking operations.
+We'll also go out of our way to explain why the usual Python async techniques don't work the way you expect in Shiny.
+
+## Using async/await in Shiny
+
+Asynchronous programming is a technique used in Python to increase scalability and responsiveness, usually in networked apps.
```
cpsievert, line 17: I don't think many readers will know what you mean by networksed

docs/reactive-nonblocking.qmd
```diff
@@ -0,0 +1,316 @@
...
+    @reactive.event(input.btn)
+    async def sum_values():
+        await asyncio.sleep(5)
+        return input.x() + input.y()
+
+    @render.text
+    async def sum():
+        return str(await sum_values())
+
+app = App(app_ui, server)
```
cpsievert, line 62: What's your reasoning for just including Core apps? Would you prefer to have a Express / Core tab option?

docs/reactive-nonblocking.qmd
```diff
@@ -0,0 +1,316 @@
...
+In this article, we'll look at how to make your Shiny apps more responsive by using non-blocking operations.
+We'll also go out of our way to explain why the usual Python async techniques don't work the way you expect in Shiny.
+
+## Using async/await in Shiny
+
+Asynchronous programming is a technique used in Python to increase scalability and responsiveness, usually in networked apps.
+Python supports async programming at the language level, using the `async`/`await` keywords; in the standard library, with the [`asyncio`](https://docs.python.org/3/library/asyncio.html) module; and in third-party libraries, like [FastAPI](https://fastapi.tiangolo.com/) and [aiohttp](https://docs.aiohttp.org/).
+
+Shiny has async support as well, but it's a bit different from your typical async Python framework.
+On the one hand, Shiny is built on top of [Starlette](https://www.starlette.io/), which is an async web framework, so it's possible to use async functions in many parts of your Shiny app. On the other hand, Shiny is also designed around reactive programming concepts, and that creates constraints on how async functions can be used.
```
cpsievert, line 21: One sentence per line, please

docs/reactive-nonblocking.qmd
```diff
@@ -0,0 +1,316 @@
...
+---
+
+Sometimes in a Shiny app, you need to perform a long-running operation, like loading a large dataset or doing an expensive computation.
+If you do this in a reactive context, it will block the rest of the application from running until the operation completes.
+This can be frustrating for users, who may think that the app has crashed.
+
+Worse, if you have multiple users, then one user's long-running operation will block the other users' apps from running as well.
+These other users will not even be aware that their slow performance is due to another user's actions.
+
+In this article, we'll look at how to make your Shiny apps more responsive by using non-blocking operations.
```
cpsievert, line 12:
```suggestion
In this article, we'll learn how to Shiny apps more responsive by using non-blocking operations.
```

docs/reactive-nonblocking.qmd
```diff
@@ -0,0 +1,316 @@
...
+We'll also go out of our way to explain why the usual Python async techniques don't work the way you expect in Shiny.
```
cpsievert, line 13: I'm not sure this is necessary. If you feel like it is, perhaps there's a less negative take on it?

docs/reactive-nonblocking.qmd
```diff
@@ -0,0 +1,316 @@
...
+Shiny supports the use of `async` and `await` in your reactive code.
+You can use `async` functions in `@reactive.effect`, `@reactive.calc`, and render functions, and within those async functions you can use `await` to wait for the results of other async functions.
+
+```{shinylive-python}
+#| standalone: true
+#| components: [editor, viewer]
+import asyncio
+import datetime
+from random import random
+import time
```
cpsievert, line 34:
```suggestion
```

docs/reactive-nonblocking.qmd
```diff
@@ -0,0 +1,316 @@
...
+
+app = App(app_ui, server)
+```
+
+However, you may be surprised to learn that this technique alone does not result in improved responsiveness in Shiny apps!
+
+In the app above, the first thing in the UI is a reactive output that displays the current time.
+Click the button and notice that, during the five seconds that the app is waiting for the (artificially slow) sum calculation to complete, the time does not update.
+Despite being defined as an asynchronous function, the sum logic is still blocking the time output.
+You could replace `await asyncio.sleep(5)` with its synchronous equivalent, `time.sleep(5)`, and you'd get exactly the same result.
```
cpsievert, line 70: I think this would read better if this came before the example

docs/reactive-nonblocking.qmd
```diff
@@ -0,0 +1,316 @@
...
+
+### When async is not async
+
+While surprising, this behavior is intentional.
+Shiny goes out of its way to ensure that reactive calcs, effects, and renderers are run in a serial fashion, never concurrently.
```
cpsievert, line 75: If you have a succinct way of explaining why that it is, it seems worth explaining.

```suggestion
Shiny goes out of its way to ensure that reactive functions are run in a serial fashion, never concurrently.
```

docs/reactive-nonblocking.qmd
```diff
@@ -0,0 +1,316 @@
...
+```
+
+However, you may be surprised to learn that this technique alone does not result in improved responsiveness in Shiny apps!
+
+In the app above, the first thing in the UI is a reactive output that displays the current time.
+Click the button and notice that, during the five seconds that the app is waiting for the (artificially slow) sum calculation to complete, the time does not update.
+Despite being defined as an asynchronous function, the sum logic is still blocking the time output.
+You could replace `await asyncio.sleep(5)` with its synchronous equivalent, `time.sleep(5)`, and you'd get exactly the same result.
+
+### When async is not async
```
cpsievert, line 72:
```suggestion
### Why does async still block?
```

docs/reactive-nonblocking.qmd
```diff
@@ -0,0 +1,316 @@
...
+
+While surprising, this behavior is intentional.
+Shiny goes out of its way to ensure that reactive calcs, effects, and renderers are run in a serial fashion, never concurrently.
+This means that if you have two (async) reactive effects that both call `await asyncio.sleep(1)`, the second one will not even begin to start executing until the first one has finished.
+This is true even if the two reactive effects belong to two different Shiny sessions (within one Python process).
```
cpsievert, line 77:
```suggestion
This is true even if the two reactive effects belong to two different Shiny sessions and one Python process is serving those sessions.
```

docs/reactive-nonblocking.qmd
```diff
@@ -0,0 +1,316 @@
...
+To create an `ExtendedTask`, use the `@reactive.extended_task` decorator on an async function.
+Let's adapt the example above to use an `ExtendedTask`:
+
+```{shinylive-python}
+#| standalone: true
+#| components: [editor, viewer]
+import asyncio
+import datetime
+from random import random
+import time
```
cpsievert, line 110:
```suggestion
```

docs/reactive-nonblocking.qmd
```diff
@@ -0,0 +1,316 @@
...
+```
+
+Note the `sum_values` function, which is where the actual (slow) work is done.
+It is decorated with `@reactive.extended_task`, which means that it will be run asynchronously and concurrently with other tasks, and that its result will be available as `sum_values.result()`.
+
+We've also added a `@ui.bind_task_button` decorator on top of the `@reactive.extended_task` decorator.
+This synchronizes the `ExtendedTask` object with the `ui.input_task_button` in the UI, so that the button will be in its "Processing..." state while the task is running.
+It _does not_ cause button clicks to invoke the task; we still need to do that manually, which we'll talk about next.
+
+If you use some other UI gesture or condition besides `ui.input_task_button` to invoke the task, don't include the `@ui.bind_task_button` decorator--it doesn't work with `ui.input_action_button`, for example.
```
cpsievert, line 153: This feels like a `::: callout-note`?

docs/reactive-nonblocking.qmd
```diff
@@ -0,0 +1,316 @@
...
+
+### Invoking an extended task
```
cpsievert, line 155:
```suggestion
### Invoking the task
```

docs/reactive-nonblocking.qmd
```diff
@@ -0,0 +1,316 @@
...
+
+Unlike a reactive effect, simply creating an extended task does not cause it to run.
+It needs to be invoked (called like a function).
+
+In this case, the `sum_values` extended task is called from the `btn_click` reactive effect (`@reactive.effect`), which runs whenever the button is clicked (`@reactive.event(input.btn)`).
+
+Notice also that the `sum_values` logic no longer reads `input.x()` and `input.y()` directly; as an extended task, it is not allowed to read reactive inputs or calcs (and attempting to do so would result in an error).
+Instead, it takes `x` and `y` as arguments, which are passed in by `btn_click`.
+
+### Using the result of an extended task
```
cpsievert, line 165:
```suggestion
### Retrieving results
```

docs/reactive-nonblocking.qmd
```diff
@@ -0,0 +1,316 @@
...
+If you use some other UI gesture or condition besides `ui.input_task_button` to invoke the task, don't include the `@ui.bind_task_button` decorator--it doesn't work with `ui.input_action_button`, for example.
+
+### Invoking an extended task
+
+Unlike a reactive effect, simply creating an extended task does not cause it to run.
+It needs to be invoked (called like a function).
+
+In this case, the `sum_values` extended task is called from the `btn_click` reactive effect (`@reactive.effect`), which runs whenever the button is clicked (`@reactive.event(input.btn)`).
+
+Notice also that the `sum_values` logic no longer reads `input.x()` and `input.y()` directly; as an extended task, it is not allowed to read reactive inputs or calcs (and attempting to do so would result in an error).
```
cpsievert, line 162:
```suggestion
Notice also that the `sum_values` logic no longer reads `input.x()` and `input.y()` directly in the function body (doing so would lead to an error). 
Instead, when invoking the task, we pass reactive values to it.
```

docs/reactive-nonblocking.qmd
```diff
@@ -0,0 +1,316 @@
...
+This is critically important for `ProcessPoolExecutor` to work correctly because of how Python's `multiprocessing` module works: only module-level functions can survive the trip to a worker Python subprocess.
+
+There's also the `app.on_shutdown(pool.shutdown)` line at the end of each example.
+This is necessary to ensure that the pool is shut down when the app stops running.
+In particular, if you're using `ProcessPoolExecutor` and neglect to shut down the pool, you can end up with orphaned Python processes hanging around.
+
+## Summary
+
+Only use async functions in your reactive code if you need to call async-only functions.
+Don't expect your application to run faster, more responsively, or more efficiently.
```
cpsievert, line 314:
```suggestion
Don't expect your application to run faster, more responsively, or more efficiently with async functions alone.
```

docs/reactive-nonblocking.qmd
```diff
@@ -0,0 +1,316 @@
...
+Notice also that the `sum_values` logic no longer reads `input.x()` and `input.y()` directly; as an extended task, it is not allowed to read reactive inputs or calcs (and attempting to do so would result in an error).
+Instead, it takes `x` and `y` as arguments, which are passed in by `btn_click`.
+
+### Using the result of an extended task
+
+The `sum` output is now a regular reactive calc, which reads `sum_values.result()` to get the result of the extended task.
+This output actually does not "wait for" the extended task to complete, exactly; instead, it will run multiple times, as the extended task goes through different states.
+For each state, `sum_values.result()` will behave differently:
+
+* **Not yet clicked:** Raises a silent exception, which will cause the `sum` output to display nothing.
```
cpsievert, line 171:
```suggestion
* **Not yet invoked:** Raises a silent exception, which will cause the `sum` output to display nothing.
```

docs/reactive-nonblocking.qmd
```diff
@@ -0,0 +1,316 @@
...
+* **Running:** Raises a special type of exception that tells Shiny to keep the output in the "in progress" state.
+* **Successfully completed:** Returns the return value of `sum_values`, in this case an integer.
+* **Completed with an exception:** If `sum_values` raised an exception while processing, then re-raise that same exception, causing it to be displayed to the user in the `sum` output.
```
cpsievert, line 174:
```suggestion
* **Completed with an exception:** If `sum_values` raised an exception while processing, then that same exception is re-raised, causing it to be displayed to the user in the `sum` output.
```

docs/reactive-nonblocking.qmd
```diff
@@ -0,0 +1,316 @@
...
+Calling `sum_values.cancel()` on a task that isn't running will have no effect.
+
+#### Multiple invocations
+
+An extended task can run concurrently to reactive code and to other extended tasks--that's its whole purpose.
+However, a single extended task object cannot run itself multiple times concurrently. If you call `sum_values()` while it is already running, it will enqueue the new invocation and run it after the first one completes.
+
+This is often not the behavior you want, especially if the task takes a long time to complete.
+A user may accidentally click an action button twice, or they may click it again because they think the first click didn't work.
+To prevent this, use replace `ui.input_action_button` with `ui.input_task_button`, which will automatically prevent subsequent clicks until the task completes.
```
cpsievert, line 195:
```suggestion
To prevent this, use `ui.input_task_button` instead of `ui.input_action_button` to invoke the task since the former automatically prevents subsequent clicks until the task completes.
```

docs/reactive-nonblocking.qmd
jcheng5: This was intended to go live before Express, it's supposed to merge into main not rc-express. Although the window keeps shrinking...

docs/reactive-nonblocking.qmd
jcheng5: I'd give anything for an auto-formatter

docs/reactive-nonblocking.qmd
jcheng5: What started this whole thing was Shiny for Python users asking why using the async keyword didn't seem to make things concurrent. I do think it's quite surprising that we're a web framework that has async, yet it doesn't do the thing you expect async to do.

I'll tweak the wording

docs/reactive-nonblocking.qmd
```diff
@@ -0,0 +1,316 @@
...
+from shiny import App, reactive, render, ui
+
+app_ui = ui.page_fluid(
+    ui.input_numeric("x", "x", value=1),
+    ui.input_numeric("y", "y", value=2),
+    ui.input_task_button("btn", "Add numbers"),
+    ui.output_text("sum"),
+)
+
+pool = concurrent.futures.ProcessPoolExecutor()
```
cpsievert, line 278: Feels like it's be better UX to just callout this change rather than listing the entire example again? Or put both examples into a tabset?

docs/reactive-nonblocking.qmd
jcheng5: I'd like to keep the tense the same as the other bullets--I'll change "re-raise" in my original to "re-raises". Is that better?

docs/reactive-nonblocking.qmd
```diff
@@ -0,0 +1,316 @@
...
+
+::: {.callout-note}
+The examples below work well in Shiny Core, but a naive port to Shiny Express will not work as well.
+The `ThreadPoolExecutor` and `ProcessPoolExecutor` objects need to be created as module-level variables, not as session-level variables, since we'd ideally like to pool resources across all sessions.
+:::
+
+::: {.callout-note}
+`ProcessPoolExecutor` is not available in Shinylive (e.g. WASM mode).
+`ThreadPoolExecutor` is available in Shinylive and appears to work, but doesn't: it actually performs all of its work on the main thread.
+:::
```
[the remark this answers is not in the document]
jcheng5: The shared module concept helps, but not enough--the pools cannot be cleaned up when the app shuts down, both because 1) there's no `app` object to call `app.on_shutdown` on, and 2) the shared module concept wouldn't give you access to the app object even if there was one.

We can definitely devise a way to do this in the future though.

docs/reactive-nonblocking.qmd
jcheng5: Good call, especially since they don't actually run in the page.

docs/reactive-nonblocking.qmd
jcheng5: I like the stronger wording. It's true that async code will perform the same as sync, and we've made it clear there is a different path you need to take if you want concurrency. I think the stronger wording is important because it's such a violation of expectations.

docs/reactive-nonblocking.qmd
cpsievert: Yep, thanks

docs/reactive-nonblocking.qmd
cpsievert: 👌

docs/reactive-nonblocking.qmd
cpsievert: We'll need to come back to this before merging rc-express (if we don't, it'll feel awkward that everything else in reactivity and before is Express first, but this is Core first) 

So, ideally, I'd prefer to do those changes now and merge into rc-express. AFAICT, there is nothing in the critical path of the content there (other than the API reference, which doesn't seem absolutely critical, and should get done this week).

cpsievert (approved):

PR #78, opened 2024-01-23
Split API docs into separate Express and Core

jcheng5:
Goes with https://github.com/posit-dev/py-shiny/pull/1053


cpsievert:
Manually cherry-picked commits into #64

PR #83, opened 2024-01-26
tooltip styling and express/core tab accentuation

gregswinehart:
@gadenbuie thanks so much for this.  

I'm wondering if we could use Bootstrap's "[dismiss on next click](https://getbootstrap.com/docs/5.3/components/popovers/#dismiss-on-next-click)" option for the popovers?


gadenbuie (approved):
🚢 🚀

PR #86, opened 2024-01-27
Expand on Express vs. Core

jcheng5:
The goal with this PR is to end up with these articles under a "Syntax modes" section:

| Filename | Title | Description |
| --- | --- | --- |
| express-vs-core.qmd | **Express vs. Core** | Describes the motivation and details the differences from an authoring perspective |
| express-or-core.qmd | **Choosing a syntax** | Gives opinionated advice as to when you'd want to use one or the other |
| express-in-depth.qmd | **Express in depth** | Explores the fuzzy frontier where Express's abstractions start to get in the way, and the workarounds we've provided where possible |
| express-to-core.qmd | **Transition to Core** | Instructions for moving from Express to Core |


_quarto.yml
```diff
@@ -237,9 +237,10 @@ website:
...
             - docs/reactive-patterns.qmd
             - docs/reactive-mutable.qmd
-        - section: "Express vs Core"
+        - section: "Syntax modes"
           contents:
-            - docs/express-introduction.qmd
+            - docs/express-vs-core.qmd
+            - docs/express-or-core.qmd
             - docs/express-in-depth.qmd
             - docs/express-to-core.qmd
```
jcheng5, line 245: I didn't intentionally make this `express-[vs|or|to]-core.qmd`, maybe revisit by end of the weekend

docs/express-or-core.qmd
```diff
@@ -0,0 +1,59 @@
...
+The most important difference between the two syntaxes is that Express allows you to intermingle UI and server code, while Core requires you to separate them.
+This can feel inconvenient while adding features to your app, as each new output requires you to jump between two different sections of the app file.
+
+But for larger and longer-lived apps, this separation of concerns becomes a considerable advantage.
+It is much easier to add, remove, or relocate pieces of your UI when all of its code is in one place, with no server code to confuse things.
+
+Take a look at the following application, written both ways:
+
+TODO
+
```
jcheng5, line 33: This article is still very much WIP

PR #93, opened 2024-01-30
Format some CSS and HTML files

cpsievert:
There are no actual changes here, just formatting applied by VS Code

### DIFF
--- a/components/_partials/components.css
+++ b/components/_partials/components.css
@@ -1,5 +1,5 @@
 /* make space for sidebar icons */
-@media only screen and (min-width:992px) {
+@media only screen and (min-width: 992px) {
   .components-hero-img {
     margin-left: -60px;
     width: 550px;
@@ -13,7 +13,7 @@
   }
 }
 
-@media only screen and (min-width:1350px) {
+@media only screen and (min-width: 1350px) {
   .components-hero-img {
     margin-left: -60px;
     width: 650px;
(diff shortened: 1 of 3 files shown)


components/_partials/components.css
```diff
@@ -77,24 +79,28 @@
...
   transition: all 0.3s !important;
 }
 
 a:hover .component-link-icon:after,
-a.component-list-header-text:hover ~ div.component-list-icon p a i.component-link-icon:after {
+a.component-list-header-text:hover
+  ~ div.component-list-icon
+  p
+  a
+  i.component-link-icon:after {
```
gadenbuie, line 103: okay, this one is a little weird

include-in-header.html
```diff
@@ -13,123 +13,145 @@
 
 <!-- Google Tag Manager -->
 <script>
-  (function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start': new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src='https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);})(window,document,'script','dataLayer','GTM-KHBDBW7');
+  (function (w, d, s, l, i) {
```
gadenbuie, line 16: I do think applying prettier to the entire file is a bit aggressive in this particular file, especially for these lines.

Can we just prettify selections instead?

gadenbuie (approved):

PR #94, opened 2024-01-30
Refactor/consolidate tabset styling, use shiny-mode-tabset more generally

cpsievert:
The main purpose of this PR is to allow `.shiny-mode-tabset` to be used more generally throughout the site (and look better without hacks on https://shiny.posit.co/py/docs/ui-overview.html) 

While I was there, I discovered we have a lot of overlapping tabset styling logic, scattered in 3 different places. This PR removes that redundancy and also leans on `.nav-underline` to provide (most of) the underline styling.

### DIFF
--- a/include-in-header.html
+++ b/include-in-header.html
@@ -84,7 +84,7 @@
       const tooltipDisplay = `<i class="bi bi-question-circle-fill d-inline d-sm-none"></i><span class="d-none d-sm-inline">What's the difference?</span>`;
 
       document
-        .querySelectorAll('.panel-tabset[data-group="shiny-app-mode"]')
+        .querySelectorAll(".panel-tabset.shiny-mode-tabset")
         .forEach(function (tabset) {
           const trigger = document.createElement("div");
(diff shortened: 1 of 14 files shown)


include-in-header.html
```diff
@@ -107,18 +107,25 @@
         });
     }
 
-    function addPanelPillsNav() {
+    function handlePanelVariants() {
```
gadenbuie, line 110: `.panel-pills` and `.panel-underline` should be core Quarto features! I think you should consider PR-ing this to quarto or at least putting this function in an issue so others can use it.

That said, you should leave a note here linking to that issue or at least a reminder that we'd want to take this out if `.panel-pills` ships in quarto.

(I could see quarto using `.panel-tabset-underline` and `.panel-tabset-pills` too, similar to their `.callout-{tip,warning,...}` classes.)

gadenbuie (approved):
LGTM, thanks for consolidating these!

cpsievert:
@gregswinehart merging this for now since this has some important fixes, but feel free to send me your CSS tweaks for the tabsets!

PR #111, opened 2024-02-26
Fix components gallery missing Variations heading

gadenbuie:
Restores the missing **Variations** heading in the components gallery.

Also switches to using CSS to add the section dividers and spacing via `border-top` on `h2` elements, except the first `h2` on the page. This works for both `components/` and `layouts/`.

I also made sure that sections in `components/` that are below `## Details` are level-3 headings (e.g. `### Another detail`) so they fit in the Details section.

While looking at the headings, I noticed that there were just heading jumps (e.g. an `<h5>` nested inside an `<h2>`). I updated the heading structure to use `h1` through `h3`. Most of the `<h5>` elements were used for styling, so I switched these out for `<div class="h5">` to get the same affect without putting the element in the accessibility tree.

While looking at that (using the [axe DevTools extension](https://www.deque.com/axe/devtools/)) I found a few images that are primarily decorative that were missing `alt=""` attributes.

### DIFF
--- a/components/_partials/components-list.ejs
+++ b/components/_partials/components-list.ejs
@@ -4,7 +4,7 @@
 <div class="g-col-md-4 g-col-sm-6 g-col-12 component-list-column">
 <div class="component-list-header">
 <a class="component-list-header-text" href="<%= item.path %>">
-<h5><%= item.title %></h5>
+<div class="h5"><%= item.title %></div>
 </a>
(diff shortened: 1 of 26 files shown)


gregswinehart:
Looks great @gadenbuie !!

gregswinehart (approved):

PR #119, opened 2024-03-05
Use lower-case `reactive.calc`, `.effect`, and `.value`

wch:
This PR updates examples to use lower-case  `reactive.calc`, `.effect`, and `.value`.


gadenbuie (approved):
🔤 LGTM!

PR #124, opened 2024-03-08
Update to shinylive 0.2.4

wch:
This updates to shinylive 0.2.4, so that it will have Shiny 0.8.1 in shinylive.


gadenbuie (approved):
Thanks!

PR #125, opened 2024-03-09
Put components in custom elements

wch:
This PR is an experiment to with putting components in custom elements.

To do:

- [x] Extract HTML dependencies from Shiny and make them available in the page for the components to use.
- [x] Add remaining components.
- [x] Update styling.

Live test page is here: https://pr-125--pyshiny.netlify.app/docs/test.html

Here's a screen shot of the test page. Note that I haven't added all components in the test page yet, and HTML dependencies (other than Bootstrap) aren't loaded, so some components don't work yet:

image


cpsievert:
I think this was done by Joe in another PR?

PR #126, opened 2024-03-11
use a template for the template listing

gadenbuie:


gadenbuie:
@cpsievert Something's just a little borked with the last section

image

PR #127, opened 2024-03-12
input task button component entry

chendaniely:


components/inputs/input-task-button/app-core.py
gadenbuie, line 1: This app is a good technical demo -- _see, it works!_ for people who already deeply understand reactivity. 

I don't think it's a good example for actually demonstrating the value of the extended task. Most people will want to use extended tasks to kick off a long-running computation without preventing users from interacting with the rest of the app. In other words, there are _fast tasks_ and _slow tasks_. 

This example suffers from two problems:

1. The fast task isn't driven by the user. In this example, the time updates, but I don't think the connection to "your app still runs" will be clear to learners.
2. The slow task isn't (pretending) to be something the learner would actually do in their app.

I have three recommendations:

1. Make sure the fast task involves the user directly interacting with the app.
1. Make the slow task longer to give the user more time to see the overlapping work.
1. Make the slow task at least appear to be something people would do, like build a model, read from a database, crunch some stats, hit an API, etc. (It doesn't have to actually do that, but it should pretend to.)

gadenbuie (commented):
Thanks Dan! I have a few comments, let me know if you have any follow-up questions or if I can help clarify anything.

chendaniely:
whoops! clicked an update button, the updated PR is coming in a bit!

PR #128, opened 2024-03-15
shinylive and format fixes for components

gadenbuie:
Fixing an issue with https://shiny.posit.co/py/components/outputs/map-ipyleaflet/#variations

image

The issue was introduced because the `app*-preview.py` file contains the contents of a quarto shinylive code chunk, which includes a JSON file. JSON is _almost but not quite_ the same structure as a Python dictionary, and `ruff` or `black` will add trailing commas, invalidating the JSON.

I've added `ruff.toml` to exclude these files, making it possible to call `ruff format` at the project root without re-introducing the issue.


gregswinehart (approved):
This went deeper than I thought! Thank you so much @gadenbuie

PR #134, opened 2024-04-10
Add input_dark_mode component

skaltman:


gregswinehart (approved):
Looks great! Thank you so much @skaltman

gadenbuie (commented):
This is a great start, thank you @skaltman!

I do think we should consider adding a variation for the very common use of `ui.input_dark_mode()` in a `ui.page_navbar()`. It would be helpful to show the pattern of using `ui.nav_spacer()` to push the dark mode switch to the right.

If you don't have bandwidth for that at the moment, we can follow up in another PR.

skaltman:
> I do think we should consider adding a variation for the very common use of `ui.input_dark_mode()` in a `ui.page_navbar()`. It would be helpful to show the pattern of using `ui.nav_spacer()` to push the dark mode switch to the right.

I can add this variation! Thanks @gadenbuie

skaltman:
@gadenbuie I added the navbar variation, but am having trouble getting it to work in Shiny Express. Do you mind writing that code, or pointing me to an example? I'm specifically confused about where to add the `ui.nav_spacer()` in a Shiny Express app.

gadenbuie:
@skaltman It turns out I recently added an example that was just released today. The example in [ui.page_opts()](https://shiny.posit.co/py/api/express/express.ui.page_opts.html) should give you a good place to start (or you probably could just copy the example from there).

skaltman:
Thanks for pointing me to that example @gadenbuie. I added the express version of the variation, and this PR is ready for review. 

I did notice that the preview version of the variation doesn't show how the spacer pushes the dark mode switch to the right, however, because of the small screen width. Let me know if there's something I should do about that.

garrettgman:
@gregswinehart and @gadenbuie just want to doublecheck if there's anything else you need on our end. I can't wait to see this on the site :)

gadenbuie:
Thanks for the nudge @garrettgman, this fell off my radar.

> the preview version of the variation doesn't show how the spacer pushes the dark mode switch to the right

I added a little feature to scale the iframe down a bit (or at least I tried, I have to wait for the preview here to see if it worked).

I also adjusted the examples to use `ui.nav_control()` instead of `ui.nav_panel()`. `nav_control()` adds a button or text to the navbar without creating a linked panel.

gadenbuie:
It works!
image

Thanks again @skaltman!

PR #140, opened 2024-04-25
Static previews for input components

jcheng5:
The preview thumbnails on the [component listing page](https://shiny.posit.co/py/components/) are py-shinylive apps. This makes them extremely slow to load. For the inputs in particular, having them as py-shinylive apps doesn't have any advantage vs. static HTML since there aren't any outputs being displayed, and the slow loading is a particular problem because they're the first components on the page that you see.

This PR allows component gallery items to opt-in to have their appPreview be shown statically, that is, compiled to static HTML under the `components/static` subdirectory. There should be no user experience difference versus before, except that the inputs load faster.

**Note:** `quarto preview` has huge slowdowns when viewing the component listing page with this branch; the page itself renders quickly, but rendering the _static_ preview .html files in iframes takes a really long time and blocks the server! This [quarto-cli PR](https://github.com/quarto-dev/quarto-cli/pull/9491) will fix the problem, but until then, loading that page during development will be extremely painful.

## Questions/TODO

- [ ] I wasn't sure when exactly it makes sense to run the make-static-previews.py script. I didn't find a natural hook in quarto so I just added it to the makefile--it happens as part of `make all` (or you can run `make static-components`). I'd welcome feedback on that.
- [x] Need to file the PR on py-shiny that sets `all_files=True` on the Bootstrap dependency.


components/make-static-previews.py
```diff
@@ -0,0 +1,145 @@
...
+        )
+    )
+
+
+def relative_to(path1: Path, path2: Path) -> Path:
+    """
+    Equivalent to `path1.relative_to(path2, walk_up=True)`, except that walk_up=True is
+    only supported from Python 3.12 on and this function works on older versions. If and
+    when we can assume Python 3.12, this function can be removed in favor of the
+    built-in method.
```
gadenbuie, line 107: We're currently building the docs in CI on 3.10, but I don't see why we couldn't bump this to 3.12.

https://github.com/posit-dev/py-shiny-site/blob/c8774a4735f875f08585011f14700cf75db21066/.github/workflows/deploy-docs.yml#L13-L14

We should also probably have a `.python-version` in this repo that matches CI anyway.

So I vote we move up to 3.12. Since that would probably happen in a separate PR, it'd be good to leave a `TODO` comment here to make it easier to find.

components/make-static-previews.py
```diff
@@ -0,0 +1,145 @@
...
+        print(f"{app_path}\n    Writing to {html_path}")
+        start_time = time.perf_counter()
+        render_static_preview(app_path, html_path, libdir)
+        end_time = time.perf_counter()
+        print(f"    Succeeded in {((end_time - start_time) * 1000):.2f}ms")
+
+    # As of this writing, there's a bug in the bootstrap dependency that ships with
+    # shiny in that it doesn't include all_files=True, so some font stuff is missing.
+    # This code works around the bug by copying the missing files from the shiny package
+    # to the static lib directory. When the bug is fixed, this code can be removed.
```
gadenbuie, line 135: Is this a bug report? 😆 If you haven't fixed this yet, I can handle the py-shiny change.

components/make-static-previews.py
```diff
@@ -0,0 +1,145 @@
...
+def render_static_preview(appfile: Path, destfile: Path, libdir: Path):
+    appfile = rootdir / appfile
+    destfile = rootdir / destfile
+
+    app_mod = importlib.import_module(
+        str(appfile.relative_to(cwd).with_suffix("")).replace("/", ".")
+    )
+    if "app_ui" not in dir(app_mod):
+        raise RuntimeError(f"app_ui not found in {appfile}")
+    app_ui = app_mod.app_ui
```
gadenbuie, line 65: TIL, this is a neat trick!

components/make-static-previews.py
```diff
@@ -0,0 +1,145 @@
...
+from typing import Generator
+
+import shiny.html_dependencies
+from _qmd import get_qmd_split
+from htmltools import Tag, head_content, tags
+
+compdir = Path(__file__).parent  # ./components
+rootdir = compdir.parent  # repository root
+cwd = Path(os.getcwd())  # current working directory
+libdir = compdir / "static" / "lib"
```
gadenbuie, line 16: Might be worth leaving a comment that this fixed directory collects dependencies from all examples in a shared folder.

gadenbuie (approved):
Looks great, the "enriched app ui" is a really elegant approach.

gadenbuie:
> **Note:** `quarto preview` has huge slowdowns when viewing the component listing page with this branch

This was very true for me too! What helped me locally was to kick off the preview but use `http-server _build` to serve the local build.

gadenbuie:
> * when exactly it makes sense to run the make-static-previews.py script. I didn't find a natural hook in quarto so I just added it to the makefile--it happens as part of `make all`

It might make sense to also ensure the static previews are up to date when you run `make site` or `make serve`

components/make-static-previews.py
gadenbuie: https://github.com/posit-dev/py-shiny/pull/1342

components/make-static-previews.py
```diff
@@ -0,0 +1,145 @@
...
+                        url
+                    };
+                };
+                """
+            )
+        )
+    )
+
+
+def relative_to(path1: Path, path2: Path) -> Path:
```
jcheng5, line 102:
```suggestion
# TODO: Remove this function once we upgrade to Python 3.12
def relative_to(path1: Path, path2: Path) -> Path:
```

components/make-static-previews.py
jcheng5:
```suggestion
# This directory will be where we put all the dependencies from all the
# static component examples, rather than forcing the browser to fetch
# multiple redundant copies from various subfolders.
libdir = compdir / "static" / "lib"
```

components/make-static-previews.py
jcheng5: Oh, great. I had no idea where that was.

components/make-static-previews.py
jcheng5: Thank you!

PR #139, opened 2024-04-24
New home page

gshotwell:
This updates the home page copy to guide the user towards installing Shiny

PR #142, opened 2024-04-30
Remove Source code positioning rule

gshotwell:
This rule was added in the Express changes, but seems to have removed the copy button from regular code chunks (e.g https://shiny.posit.co/py/docs/comp-r-shiny.html#r)


gadenbuie:
Why'd you close this? I think it's a good idea to do this separately from #139 where it's easy to test and will merge faster.

PR #143, opened 2024-04-30
Update data grid/table row selection docs

wch:
This fixes documentation and examples on doing row selections, for Shiny 0.9.0.

### DIFF
--- a/components/outputs/data-grid/app-variation-select-rows-express.py
+++ b/components/outputs/data-grid/app-variation-select-rows-express.py
@@ -9,11 +9,11 @@ ui.h2("Palmer Penguins")
 
 @render.ui()
 def rows():
-    rows = input.penguins_df_selected_rows()  # <<
+    rows = penguins_df.input_cell_selection()["rows"]  # <<
     selected = ", ".join(str(i) for i in sorted(rows)) if rows else "None"
     return f"Rows selected: {selected}"
 
 
 @render.data_frame
 def penguins_df():
-    return render.DataGrid(penguins, row_selection_mode="multiple")  # <<
+    return render.DataGrid(penguins, selection_mode="rows")  # <<
--- a/components/outputs/data-grid/index.qmd
+++ b/components/outputs/data-grid/index.qmd
@@ -45,9 +45,10 @@ listing:
     dir: components/outputs/data-grid/
   contents:
   - title: Select Rows
-    description: Set `row_selection_mode` in `render.DataGrid()` to `"single"` to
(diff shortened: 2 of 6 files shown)


components/outputs/data-grid/app-variation-select-rows-express.py
```diff
@@ -9,11 +9,11 @@
 
 @render.ui()
 def rows():
-    rows = input.penguins_df_selected_rows()  # <<
+    rows = penguins_df.input_cell_selection()["rows"]  # <<
```
gadenbuie, line 12: I think this is a good change in the sense of having the existing example run correctly with shiny 0.9.0.

We don't necessarily need to do it in this PR, but this example is less motivating now that the output object has more methods. It'd be nice if this example or a variant in this section gave more instruction about those methods and how to use them.

components/outputs/data-grid/index.qmd
```diff
@@ -126,13 +127,13 @@ To make a reactive Data Grid, follow three steps:
...
+A Data Grid can also collect input from the user. To allow this, set `render.DataGrid(selection_mode="row")` or `render.DataGrid(selection_mode="rows")` to allow the user to select one or more rows of the Data Grid.
 
-The indices of the selected rows will be accessible within the server function as a reactive variable returned by `input.<id>_selected_rows()`, where <id> is the id of the `ui.output_data_frame()` associated with the table.
+The indices of the selected rows will be accessible within the server function as a reactive variable returned by `<name>.input_cell_selection()["rows"]`, where <name> is the name of the function decorated with `@render.data_frame`.
 
-The value returned will be `None` if no rows are selected, or a tuple of integers representing the indices of the selected rows. To filter a pandas data frame down to the selected rows, use `df.iloc[list(input.<id>_selected_rows())]`.
+The value returned will be an empty tuple if no rows are selected, or a tuple of integers representing the indices of the selected rows. To filter a pandas data frame down to the selected rows, use `df.iloc[list(<name>.input_cell_selection()["rows"])]`.
```
gadenbuie, line 136: At the very least, I think this section should give the user a hint that `.input_cell_selection()` isn't the only method (maybe in a paragraph below this one) with a pointer to our docs where they might learn more.

gadenbuie (approved):

components/outputs/data-grid/index.qmd
wch: Good idea -- I've added a bit of info and link to the API docs.

gadenbuie:
@wch I just went looking for this example and realized that the docs aren't live yet, so I'm going to merge this now.

PR #146, opened 2024-05-15
Change install instructions to work with shiny 0.9

gshotwell:
Currently the shiny create commands on the website do not work, so we should change this to a released template.

### DIFF
--- a/index.qmd
+++ b/index.qmd
@@ -143,7 +143,7 @@ pip install shiny
 :::{.code_container .homepage-code .w-100}
 ```{python}
 #| eval: false
-shiny create --template dashboard-tips
+shiny create --template dashboard
 ```
 :::
 :::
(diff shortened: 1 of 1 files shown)


gadenbuie:
> Currently the shiny create commands on the website do not work

Are you sure? I just tried `--template dashboard-tips` with a fresh `pip install shiny` and it worked as expected and created the tipping dashboard.

gshotwell:
It looks like the source of this issue is shell cache. To reproduce:
- Install old version of shiny
- `pip install shiny -U`
- Do not restart terminal or clear the shell cache

You'll see the the old click option commands in the CLI. When you restart the terminal you'll see the new ones. 

So overall that doesn't seem like an issue for the users on the front page because they won't have old versions of Shiny installed, so I'll close this PR.

PR #147, opened 2024-05-19
Update app-two-cards-express.py

bradfordjohnson:
Update module import from "shiny" to "shiny.express".

Fixes the following error when running the example: `RuntimeError: module 'shiny.ui' has no attribute 'page_opts'`


gadenbuie (approved):
Thanks @bradfordjohnson!

PR #151, opened 2024-05-24
reactive-mutable: tweak problem/solution examples

gadenbuie:
Updates the problem/solution examples in the `reative-mutable` article.

1. I put the problem/solution examples in callouts so that it's visually clear which one provides the pattern to follow.
2. Small edit to make it easier to see the problem with the leaky changes example
3. Use `@render.code` instead of `@render.text` and change "raw values" to "user values" to make it easier to read.
4. Added a section on python operations that create copies.

Preview: https://pr-151--pyshiny.netlify.app/docs/reactive-mutable.html


wch (approved):
Looks good!

PR #157, opened 2024-06-13
Enable fast mode

jcheng5:
This has two effects: making the build faster, as quarto render currently uses a slow Lua JSON parser to deserialize objects.json on every page render; and generating an objects.txt file, which can be used by sphinx so that downstream projects can interlink with us.

### DIFF
--- a/_quarto.yml
+++ b/_quarto.yml
@@ -40,6 +40,7 @@ filters:
   - interlinks
 
 interlinks:
+  fast: true
   sources:
     python:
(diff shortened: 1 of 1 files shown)


machow (approved):
LGTM--I didn't look closely, but the `fast:` flag should be stable, and will likely become the default soonish

jcheng5:
The site successfully built with API docs in place, that's good enough for me!

PR #158, opened 2024-06-14
Upgrade interlinks

jcheng5:
We would like to use interlinks' `fast: true` mode so that objects.txt becomes available.

### DIFF
--- a/_extensions/machow/interlinks/_extension.yml
+++ b/_extensions/machow/interlinks/_extension.yml
@@ -1,6 +1,6 @@
 title: Interlinks
 author: Michael Chow
-version: 1.0.0
+version: 1.1.0
 quarto-required: ">=1.2.0"
 contributes:
   filters:
--- a/_extensions/machow/interlinks/interlinks.lua
+++ b/_extensions/machow/interlinks/interlinks.lua
@@ -1,19 +1,86 @@
-local function read_json(filename)
+local function read_inv_text(filename)
+    -- read file
     local file = io.open(filename, "r")
     if file == nil then
         return nil
     end
     local str = file:read("a")
     file:close()
-    return quarto.json.decode(str)
+
+
+    local project = str:match("# Project: (%S+)")
+    local version = str:match("# Version: (%S+)")
+
+    local data = {project = project, version = version, items = {}}
+
+    local ptn_data =
+        "^" ..
+        "(.-)%s+" ..        -- name
+        "([%S:]-):" ..      -- domain
+        "([%S]+)%s+" ..     -- role
+        "(%-?%d+)%s+" ..     -- priority
+        "(%S*)%s+" ..       -- uri
+        "(.-)\r?$"         -- dispname
+
+
+    -- Iterate through each line in the file content
+    for line in str:gmatch("[^\r\n]+") do
+        if not line:match("^#") then
+            -- Match each line against the pattern
+            local name, domain, role, priority, uri, dispName = line:match(ptn_data)
+
+            -- if name is nil, raise an error
+            if name == nil then
+                error("Error parsing line: " .. line)
+            end
+
+            data.items[#data.items + 1] = {
+                name = name,
+                domain = domain,
+                role = role,
+                priority = priority,
+                uri = uri,
+                dispName = dispName
+            }
+        end
+    end
+    return data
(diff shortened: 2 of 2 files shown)


schloerke:
Closing this — but not because it doesn't work. It does, and here's the measurement, since the numbers are useful to have on record if we ever want to revisit.

### Status: the filter half of this PR already landed

`_extensions/machow/interlinks/_extension.yml` on `main` is already at `1.1.0`, and `interlinks.lua` already contains `read_inv_text` / `read_inv_text_or_json` (it prefers `<base>.txt` and falls back to `<base>.json`). `_quarto.yml` also already sets `interlinks: fast: true`. So there's nothing left in this diff to merge.

The reason fast mode is nevertheless **inert** today: the inventories are generated in the `py-shiny` submodule using `py-shiny/docs/_quarto.yml`, which has no `fast: true` — quartodoc reads that key from the config it's handed ([`autosummary.py:804`](https://github.com/machow/quartodoc/blob/main/quartodoc/autosummary.py)). So quartodoc emits `objects.json` + `_inv/*_objects.json`, and the `.txt` branch of the filter never fires.

### It works, and it's ~6x faster to parse

Instrumented the `Meta` handler with `os.clock()` under Quarto 1.9.36. That handler runs **once per rendered document** (once per pandoc process), so this is pure per-page overhead. The site loads 21,589 items / 3.54 MB of JSON per page (`objects.json` 739 KB / 2,272 items + `_inv/python_objects.json` 2.80 MB / 19,317 items; as text those are 302 KB and 1.31 MB).

| inventory format | per-document load |
| --- | --- |
| JSON (`quarto.json.decode`) | **288 ms** (mean of 4 runs) |
| TXT (Lua pattern parse) | **49 ms** (mean of 6 runs) |

→ **5.9x faster, ~239 ms/doc saved.**

End-to-end control, `quarto render templates` (24 pandoc runs, no shinylive so the delta isn't buried under WASM work), alternating modes:

| mode | run 1 | run 2 | mean |
| --- | --- | --- | --- |
| JSON | 110.10 s | 105.03 s | 107.6 s |
| TXT | 102.01 s | 95.73 s | 98.9 s |

→ **−8.7 s over 24 docs = 362 ms/doc, −8.1 % wall clock.** (Higher than the isolated number; 239 ms is the conservative floor.)

**Output is unchanged.** A scratch page exercising ``[](`x`)``, `~shortened`, `:func:`, `:class:`, `:external+python:py:class:`, and a deliberately-broken ref produced **byte-identical** `href`s in both modes (`diff` clean). `read_inv_text` parsed all 21,589 entries with no errors.

### Why we're closing anyway

Extrapolated over the 551 rendered pages:

| build | saving |
| --- | --- |
| `make site` (serial, ~35 min) | 2.2–3.3 min (6–9 %) |
| `make site-parallel` (`SHARDS=6`, ~9 min) | ~22–33 s |
| CI (10 shards, ~9 min critical path) | **~13–20 s per shard** |

This mattered a lot when a full render was one serial ~35-minute job. Now that CI renders in 10 parallel shards, the win collapses to ~13–20 s per shard, and getting it would mean either a cross-repo change (`fast: true` in `py-shiny/docs/_quarto.yml` breaks the three `mv objects.json _objects_*.json` steps in `py-shiny/docs/Makefile` plus `_combine_objects_json.py`, so: py-shiny PR + submodule bump + site `Makefile`/`_quarto.yml`/`.gitignore` updates) or a JSON→text conversion step bolted onto the site `Makefile`. Neither is worth ~15 s of shard time.

One thing to note for the future: this PR's stated motivation was "so that `objects.txt` becomes available" — i.e. publishing an inventory that external quartodoc sites can consume via `inv: objects.txt`, the way `py-shiny/docs/_quarto.yml` already does for `brand-yml`. If that need ever comes up concretely, it's a better reason to revisit this than render speed is, and it can be done without touching the submodule.

### Also ruled out, so nobody chases it

`lookup()` linear-scans all 21,589 inventory items per reference — measured at 0.78 ms/lookup (92 lookups = 72.6 ms on `api/core/MiscTypes.qmd`). With 4,424 interlink references across 403 files that's **~3.5 s site-wide**. A name-keyed index would be ~30x on that, but it buys seconds, not minutes.

PR #163, opened 2024-06-20
fix: Correct the duplicate R function name

devpowerplatform:
there are two 'observeEvent" entries. I think one should be enough right? Any reason to keep both?

### DIFF
--- a/docs/comp-r-shiny.qmd
+++ b/docs/comp-r-shiny.qmd
@@ -331,7 +331,6 @@ For the most part you can follow this naming pattern to find the function you're
 
 | R Function                             | Python Equivalent                         |
 |-----------------------------------|-------------------------------------|
-| `observeEvent`                         | `@reactive.effect`                        |
 | `reactive`                             | `@reactive.calc`                          |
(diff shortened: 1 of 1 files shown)


gadenbuie:
Good catch! The intention was likely for the first entry to be `observe` and not `observeEvent`.

docs/comp-r-shiny.qmd
```diff
@@ -331,7 +331,6 @@ For the most part you can follow this naming pattern to find the function you're
 
 | R Function                             | Python Equivalent                         |
 |-----------------------------------|-------------------------------------|
-| `observeEvent`                         | `@reactive.effect`                        |
 | `reactive`                             | `@reactive.calc`                          |
 | `bindEvent`                            | `@ractive.event`                          |
```
jcheng5, line 335:
```suggestion
| `observe`                              | `@reactive.effect`                        |
| `reactive`                             | `@reactive.calc`                          |
| `bindEvent`                            | `@reactive.event`                         |
```

schloerke:
Thank you @devpowerplatform !

schloerke (approved):

PR #174, opened 2024-07-15
Fix link to shinylive examples

cpsievert:
Closes https://github.com/posit-dev/py-shiny/issues/1533

### DIFF
--- a/docs/overview.qmd
+++ b/docs/overview.qmd
@@ -539,7 +539,7 @@ shiny create -t dashboard
 ## Development workflow
 
 See the [workflow](install-create-run.qmd) section for more information developing Shiny apps locally.
-Also keep in mind you can develop apps in the browser using the [playground](/playground).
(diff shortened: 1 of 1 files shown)


docs/overview.qmd
```diff
@@ -539,7 +539,7 @@ shiny create -t dashboard
 ## Development workflow
 
 See the [workflow](install-create-run.qmd) section for more information developing Shiny apps locally.
-Also keep in mind you can develop apps in the browser using the [playground](/playground).
+Also keep in mind you can develop apps in the browser using the [playground](/examples).
```
gadenbuie, line 542: I don't think we have a redirect in place for this

```suggestion
Also keep in mind you can develop apps in the browser using the [playground](https://shinylive.io/py/examples/).
```

gadenbuie (approved):

PR #188, opened 2024-07-25
Test whether we can remove altair soft_dependencies hack

cpsievert:
Should no longer be needed after https://github.com/posit-dev/py-shinylive/pull/36

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 jupyter
 jupyter_client < 8.0.0
 tabulate
-shinylive==0.5.0
+shinylive @ git+https://github.com/posit-dev/py-shinylive@main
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -1,7 +1,7 @@
 jupyter
 jupyter_client < 8.0.0
 tabulate
-shinylive==0.5.0
+shinylive @ git+https://github.com/posit-dev/py-shinylive@main
```
cpsievert, line 4: Don't merge until new release happens

```suggestion
shinylive==0.5.1
```

wch: The new version of shinylive is on PyPI so I've made this change.

wch (approved):

Issue #187, opened 2024-07-24
Turn off "edit this page" under `api/`


gadenbuie:
Pages under `api/` are generated at build time, so "Edit this page" links to a missing file. 

E.g. https://shiny.posit.co/py/api/express/express.ui.layout_columns.html

Sherwin-14:
I am willing to take this up can you assign me this one?

PR #189, opened 2024-07-25
Update /api/core/index.qmd

joesho112358:
Fix attempt for #187

### DIFF
--- a/api/core/index.qmd
+++ b/api/core/index.qmd
@@ -1,3 +1,7 @@
+---
+repo-actions: false
+---
+
 # Shiny Core API
 
 This page outlines Shiny _Core_'s API reference.


api/core/index.qmd
```diff
@@ -1,3 +1,7 @@
+---
+repo-actions: false
```
gadenbuie, line 2: Thanks for trying this! This will work for this page, but not for any others under `api/`. I think the solution will be to place this setting in `api/_metadata.yml` file so that it applies to the whole subtree.

The likely challenge with that approach is that we programmatically build the files under `api/` via quartodoc and there's a good chance that we clean out `api/` entirely with each build. So I'm not sure where that will leave us.

joesho112358: looks like putting it in a meta file should do it:
image

PR #196, opened 2024-08-04
Removed the output decorator from the comparison page 

Sherwin-14:
I've removed the output decorator from the comparison page, please take a look at it.

### DIFF
--- a/docs/comp-r-shiny.qmd
+++ b/docs/comp-r-shiny.qmd
@@ -79,7 +79,6 @@ While R doesn't have an exact analog to decorators they are similar to [function
 ::: callout-tip
 ### Takeaways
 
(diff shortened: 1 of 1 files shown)


docs/comp-r-shiny.qmd
```diff
@@ -177,8 +175,7 @@ app_ui = ui.page_fluid(
     ui.output_text_verbatim("txt"),
 )
 
-def server(input, output, session):
-    @output
+def server(input, output, session):
```
wch, line 178: Looks like some trailing whitespace got added here:

```suggestion
def server(input, output, session):
```

wch (approved):
Thanks for the pull request!

PR #198, opened 2024-08-05
Install quarto before installing R dependencies

wch:
We have a `Run quarto-dev/quarto-actions/setup@v2` step with the version specified. However, the `r-lib/actions/setup-r-dependencies@v2` step in the workflow was indirectly causing another copy of `Run quarto-dev/quarto-actions/setup@v2` to run, but without a version specified. This installation failed because the system doesn't want to downgrade the .deb package.

https://github.com/posit-dev/py-shiny-site/actions/runs/10250692524/job/28356982088#step:9:2457


schloerke:
Passed install and build. Merging

schloerke (approved):

PR #207, opened 2024-09-02
Attempted to fix the broken link on custom components page 

Sherwin-14:
Is there anything else needed to be taken care of?


cpsievert:
Thanks!

PR #209, opened 2024-09-05
Update `shiny create` calls for Shiny v1.1.0

gadenbuie:
Follow up after shiny v1.1.0 release:

1. Use `shiny create --github` in more places
2. Use full `--template` and `--github` and `--mode` flags. For one, it's easier to understand, for another the consistency makes it easier for us to update in the future.
3. Updated links in all templates files to point to the `.qmd` targets
4. Generally reviewed all uses of `shiny create`.

One thing to call out: the `shiny create` commands now assume shiny v1.1.0 but I wasn't sure how to call this out without getting into awkward nesting of tabs. Our general approach is to have the docs page follow the latest release, but maybe this should be called out somewhere? I'm open to suggestions.


cpsievert (approved):
Thanks!

PR #219, opened 2024-10-02
Add `.update_cell_selection()` variation to DataGrid and DataTable components

skaltman:


cpsievert:
This is great, thanks, and sorry it went unnoticed for so long!!

PR #220, opened 2024-10-03
Add event handling variation to Plotly component

skaltman:


cpsievert:
This is great, thanks and sorry it went unnoticed for so long!!

PR #224, opened 2024-10-11
Fix incorrect package

wch:
This fixes an incorrect package. In #31, the package in requirements.txt was added as `favicons`, but it should be `faicons`.


wch:
The build is failing; I believe it's due to some change in the most recent version of quartodoc.

schloerke (approved):

PR #251, opened 2025-02-25
quarto: Update to v1.6.42

gadenbuie:


chendaniely:
assuming the workflow_dispatch was added to trigger the build. this adding this might also help next time the workflow is updated:

```yaml
on:
  push:
    branches: [main]
  pull_request:
  workflow_dispatch:
  paths:
    - .github/workflows/deploy-docs.yml

```

PR #256, opened 2025-03-05
Improved sidebar styling for 'Learn Shiny'

cpsievert:
**Before**

Screenshot 2025-03-05 at 10 19 05 AM

**After**

Screenshot 2025-03-05 at 10 18 34 AM

### DIFF
--- a/_quarto.yml
+++ b/_quarto.yml
@@ -232,49 +232,49 @@ website:
       collapse-level: 2
       align: left
       contents:
-        - section: "Essentials"
+        - section: "<span class='emoji-icon'>📌</span>  __Essentials__"
           contents:
             - docs/overview.qmd
             - docs/user-interfaces.qmd
-        - section: "Workflow"
+        - section: "<span class='emoji-icon'>💻</span>  __Workflow__"
           contents:
             - docs/install-create-run.qmd
             - docs/debug.qmd
-        - section: "User interfaces"
+        - section: "<span class='emoji-icon'>🎨</span>  __User interfaces__"
           contents:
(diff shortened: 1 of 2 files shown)


_quarto.yml
```diff
@@ -232,49 +232,49 @@ website:
       collapse-level: 2
       align: left
       contents:
-        - section: "Essentials"
+        - section: "<span class='emoji-icon'>📌</span>  __Essentials__"
```
gadenbuie, line 235: Did you consider adding bold with CSS? This seems like a reasonable change that would be useful elsewhere, even if the icon isn't included.

gadenbuie (approved):
I like this change a lot! My one nit is that It's now a little harder to find which page is currently open. It might be worth thinking about a visual treatment we could add to the active sidebar link

_quarto.yml
cpsievert: The `display: flex;` layout only works as intended if both the emoji and text are wrapped in an HTML element.

gadenbuie:
> Agreed, how about this? FWIW, the components/layouts sidebar have the same issue, but it's maybe less of an issue since there are only a few sections, and each section has many items

I like it! Another idea that would work everywhere is something like this:

```css
.sidebar-link.active::before {
    content: "\23F5";
    position: absolute;
    left: -0.9em;
    font-size: 1em;
    color: var(--bs-primary);
}

.sidebar-link.active {
    position: relative;
}
```

Which puts a carat next to the selected item. That works on the components/layouts pages too.

image

PR #258, opened 2025-03-06
Import render from shiny.express

### LINKED ISSUE #110: Express examples in Component gallery should use `from shiny.express import render`
The Express examples in the Component gallery use `from shiny import render`, but they should use `from shiny.express import render`

For example:

https://shiny.posit.co/py/components/outputs/plot-matplotlib/

image

cpsievert:
Closes #110


gadenbuie (approved):
Looks good! I think you need to run [make components](https://github.com/posit-dev/py-shiny-site/blob/5a76f4b5c9bc4c73f13f43f886c5fa9b555d83d3/Makefile#L117) before merging to at least update the shinylive links.

PR #260, opened 2025-03-11
Add kitchen sink examples to components detail pages

gregswinehart:
This PR adds @karangattu's work on "kitchen sink" examples for components: A single, catch-all app to show all (or most) of the variations of a component (eg. action button) in our documentation. This PR primarily focuses on Shiny's input components.


karangattu (approved):
LGTM :shipit:

PR #268, opened 2025-04-08
chore: typo in debug.qmd sanitize error message

chendaniely:
The default is sanitize with value of `True`. to unsanitize, set to `False`


gadenbuie (approved):
Thanks!

PR #281, opened 2025-04-24
chore: remove ?_gl* in URLs

chendaniely:
Removing the `?_gl*` parts to links.

### DIFF
--- a/docs/debug.qmd
+++ b/docs/debug.qmd
@@ -123,12 +123,12 @@ Note also that Shiny applications use Python's asyncio under the hood, so it may
 
 ### shinyapps.io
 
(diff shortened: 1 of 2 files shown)


gadenbuie:
Do you know why it's there in the first place?

chendaniely:
no idea on the why. When Joe took a look at it on the [chatbot page](https://shiny.posit.co/py/docs/genai-chatbots.html) it he got to a page where that was injected into the browser while keeping the original URL as a separate attribute.

I'm unable to reproduce that URL mangling.

I'm not sure if this is related to larger website analytics tracking?
And looking at the blame, I don't think the URL was indented to be put into the qmd document that way.

image

schloerke:
> Do you know why it's there in the first place?

From IRL convos...

Greg did not know about the app link in the yaml file.

Carson added the qmd links.

----------------

☠️ all the trackers!

schloerke (approved):

PR #282, opened 2025-04-25
chore: fix broken URLS

chendaniely:
Fix the broken URLs reported in this CI run: https://github.com/posit-dev/py-shiny-site/actions/runs/14605882422
that uses the lychee link checking in #270 

The support.posit.co URLs are ignored.

For the links and pages I fixed, I also updated the URLs to point to the core or express pages without relying on redirect.


cpsievert:
Thanks!

PR #283, opened 2025-04-25
fix: Hero text

gadenbuie:
Updates the hero text on the front page to use Bootstraps `.display-1` for `font-size` and `line-height`, resulting in better display at all screen sizes.

This also untangles the `.gradient-text` class from font size and weight considerations.

## Before

image

## After 

image


cpsievert (approved):
Thanks!

PR #286, opened 2025-04-29
chore: Improve get started app example

chendaniely:
Change the initial get started app to how a more data centric example.


chendaniely:
before `aes` move:

Kapture 2025-04-29 at 15 36 38

after `aes` move:

Kapture 2025-04-29 at 15 56 50

cpsievert:
I very much like this direction, but one small nit. I think there is room to simplify the code a bit. In particular, I think you could get rid of the `height=300` part?

gadenbuie:
> but one small nit. I think there is room to simplify the code a bit. In particular, I think you could get rid of the `height=300` part?

We could but the plot doesn't look as good and neither does the app (we don't want it to scroll in that container)

chendaniely:
this is what it looks like without the height specification.

Totally separate conversation would be to see if render.plot() can automatically fill the container it's in so it won't scroll

Kapture 2025-04-30 at 14 48 13

cpsievert:
Oh, I was assuming the code in the code block wasn't getting used to generate the result. In that case, nevermind.

chendaniely:
should we merge this? and then link to a deployed version as a separate PR? or should i deploy this version and update the site to the deployed app to help with load times?

gadenbuie (approved):
Looks great, thanks! Let's merge this and handle deploying the demo app separately

PR #288, opened 2025-04-30
Don't show a flash of shinylive source code

cpsievert:
Perhaps we should do this more generally too since the components pages suffer from the same problem.

And also seems like shinylive should be doing this automatically?

### DIFF
--- a/components/_partials/components.css
+++ b/components/_partials/components.css
@@ -80,3 +80,7 @@ main a.nav-link:hover {
   height: calc(1 / var(--scale-factor) * 100%);
   transform-origin: top left;
 }
+
+.shinylive-python code {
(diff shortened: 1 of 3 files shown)


gadenbuie:
I think shinylive should handle this but it will take some thought. In the mean time, we should also do this in `get-started/index.qmd`.

components/_partials/components.css
```diff
@@ -80,3 +80,8 @@ main a.nav-link:hover {
   height: calc(1 / var(--scale-factor) * 100%);
   transform-origin: top left;
 }
+
+.shinylive-python code {
```
cpsievert, line 84:
```suggestion
.shinylive-wrapper code {
```

PR #314, opened 2025-07-24
Add login/logout app example with dynamic UI

karangattu:
Introduced a new Shiny app demonstrating login and logout functionality using dynamic UI rendering based on authentication state. Updated index.qmd to document and showcase the new example, including preview and code links.


karangattu:
- [ ] Add intent test

PR #364, opened 2026-03-27
feat: Add llms-full.txt

### LINKED ISSUE #363: feat: Add llms.txt
We should add an llms.txt
Based on inspect.aisi.org.uk/llms.txt

We should also add some sort of action or process to make sure it gets updated when we add new things to the site.

elnelson575:
Closes #363

Auto-generated `llms-full.txt` following the [llmstxt.org](https://llmstxt.org/) spec, so LLMs can discover and consume Shiny for Python documentation.

## Key decisions

- **`llms-full.txt` (572 KB full content)** — includes cleaned page content for RAG/coding assistant use, matching the pattern used by Pydantic and others
- **Content scope:** Get Started, Concepts, Components (with examples), Layouts, API Reference, Templates. Gallery excluded (most of these are links to other repos or content made by others)
- **Deterministic generation** — output is a pure function of source files; only changed pages produce diffs
- **CI freshness check** — build fails with instructions if files are out of date
- **`_quarto.yml` is the source of truth** for site structure and page ordering; templates discovered by directory scan (not in sidebars)
- **QMD cleaning** strips frontmatter, Quarto directives, raw HTML, shinylive metadata; preserves prose and code blocks
- **Page descriptions** use title only (86/119 pages lack descriptions in frontmatter)
- **`@dataclass` for site structure types** — auto-generates `__init__`, `__repr__`, and structural `__eq__`; the last is essential for test assertions like `assert entries == [SidebarEntry(...)]`. `NamedTuple` was the alternative but feels semantically odd for types with mutable `list` fields.

## Usage

```bash
make llms-full-txt          # generate both files
make all               # includes llms-txt in full build
```

## Test plan

- [x] 38 unit tests pass (`python3 -m pytest tests/test_generate_llms_txt.py`)
- [x] Verify `llms-full.txt` has cleaned content (no `:::`, no frontmatter, no raw HTML)
- [x] Verify determinism (run twice, diff is empty)
- [x] Verify CI check catches stale files (https://github.com/posit-dev/py-shiny-site/actions/runs/23818115070/job/69423076433?pr=364)

---

<details>
<summary>Implementation plan</summary>

## File Structure

| File | Action | Responsibility |
|---|---|---|
| `scripts/generate_llms_txt.py` | Create | Main generation script — all logic lives here |
| `tests/test_generate_llms_txt.py` | Create | Unit tests for parsing and cleaning functions |
| `Makefile` | Modify | Add target, update `all` |
| `_quarto.yml` | Modify | Add`llms-full.txt` to resources (`llms.txt` is generated live from quarto from previous PR |
| `.github/workflows/deploy-docs.yml` | Modify | Add CI freshness check |
| `llms-full.txt` | Create (generated) | Full content file |
</details>


.github/workflows/deploy-docs.yml
```diff
@@ -55,6 +55,16 @@ jobs:
...
 
+      - name: Verify llms-full.txt is up to date
+        run: |
+          make llms-full-txt QUARTO_PATH=quarto
+          if ! git diff --exit-code llms-full.txt; then
+            echo ""
+            echo "ERROR: llms-full.txt is out of date."
+            echo "Run 'make llms-full-txt' locally and commit the updated file."
+            exit 1
+          fi
```
cpsievert, line 62: I think it probably makes sense for this to come earlier in the workflow. In particular, `make all` takes a long time -- It would be frustrating to wait for the site to be built only for it to err.

tests/test_generate_llms_txt.py
```diff
@@ -0,0 +1,613 @@
+import sys
```
cpsievert, line 1: It'd be helpful to have a docstring at the top of the file here to explain the intention behind the tests, when to run them, etc.

Makefile
```diff
@@ -131,6 +131,13 @@ components-static: $(PYBIN) deps
 components-shinylive-links: $(PYBIN) deps
 	. $(PYBIN)/activate && python components/update-shinylive-links.py
 
+
+## Generate llms-full.txt
+.PHONY: llms-full-txt
+llms-full-txt: $(PYBIN) quartodoc
+	. $(PYBIN)/activate && python scripts/generate_llms_txt.py
```
cpsievert, line 138: Do we not also want to run the tests after generation?

.github/workflows/deploy-docs.yml
```diff
@@ -55,6 +55,16 @@ jobs:
         run: |
           make all QUARTO_PATH=quarto
 
+      - name: Verify llms-full.txt is up to date
+        run: |
+          make llms-full-txt QUARTO_PATH=quarto
+          if ! git diff --exit-code llms-full.txt; then
+            echo ""
+            echo "ERROR: llms-full.txt is out of date."
+            echo "Run 'make llms-full-txt' locally and commit the updated file."
```
gadenbuie, line 60: To me, `llms-full.txt` seems like more of a build artifact than a source artifact. I'd personally prefer that it be built at render time and not checked into source.

It's not that I don't see the utility in tracking changes. If we were using a `gh-pages` branch, we'd have a copy of changes, which is nice.

But I have a strong aversion to having CI fail for things that could be automated. I have come to accept that we do this for build assets for _reasons_ but it's not a pattern I'd recommend we follow for this file in particular.

llms-full.txt
```diff
@@ -0,0 +1,14315 @@
...
+An action button appears as a button and has a value that increments each time the user presses the button.
+
+See also: [Action Link](https://shiny.posit.co/py/components/inputs/action-link/)
+
+---
+#### Action Link
+
+An action link appears as a link in your app and has a value that increments each time the user presses the link.
+
+See also: [Action Button](https://shiny.posit.co/py/components/inputs/action-button/)
```
gadenbuie, line 27: This whole section surprised me a bit. I would have expected a section like this to include code snippets showing the actual code to create the inputs.

Also the links seem reasonable, but they point to non-LLM-oriented pages. If we're making this whole system of documentation easier to navigate for LLMs, I'd think we'd prefer to link to LLM-ready pages.

I'm not saying that we **have to have** an LLM-ready markdown file for `py/components/inputs/action-button/`, but given how little other information is included in this section, we're basically telling the model to go read that file. So there's an asymmetry here.

scripts/generate_llms_txt.py
```diff
@@ -0,0 +1,618 @@
...
+    return re.sub(r"\[([^\]]+)\]\(([^)]+)\)", replace, content)
+
+
+_TEMPLATE_HEADINGS = re.compile(
+    r"^#{1,3} (Details|Variations|Relevant Functions)$"
+)
+
+
+def clean_qmd_content(content: str, file_path: str = "") -> str:
+    """Strip Quarto-specific markup from .qmd content, keeping prose and code."""
```
gadenbuie, line 81: Did you look into using `pandoc` to handle this conversion? Quarto's markdown is really just pandoc markdown, and pandoc can handle moving between markdown variants.

It's more work, but I think much of this could be done through a Lua filter if the goal is to suppress Quarto-specific elements. Just in general, that stack is much better set up to handle parsing and rewriting markdown.

Frankly, I'm confident this function works _right now_ but I wouldn't want to maintain it at all. Claude changes the calculus and all, but my larger worry than this particular function is that this code blends low-importance mechanical transformations with high-level assumptions about how we've laid out our docs, etc.

Relatedly, did you look into using something like [markdownify](https://pypi.org/project/markdownify/) or [markdownify-rs](https://pypi.org/project/markdownify-rs/)? They work on HTML, but if this code moved away from dealing with the low-level details of parsing our source content I think it'd be more maintainable for us in the long run (not to mention more repeatable if we do something like this in other repos).

---

At one more level of abstraction higher, if we feel that this overall process is important I think it's worth looking into how to do this inside of Quarto. This kind of extension would be useful anywhere else in the ecosystem, like in quartodoc or Great Docs, and it'd probably be a lot easier to write and maintain if it were closer to the Quarto toolchain (i.e. pandoc and Lua filters).

OTOH, if we're just trying to get this done it seems like using something else that handles the HTML parsing for us would be beneficial even if the output isn't as tight as we'd like.

PR #367, opened 2026-04-03
Chore: Upgrade to Quarto 1.9.36

elnelson575:
## Summary
- Add a Compare Versions command for regression testing
- Upgrade Quarto 1.7.23 → 1.9.36
- Enable `llms-txt` for LLM-friendly site output
- Migrate Plausible analytics to native `plausible-analytics` config

Claude's analysis based on the Quarto CHANGELOG:
| Version | Breaking Change | Affects us? |
| :--- | :--- | :--- |
| 1.7.x | Custom pandoc.tex template changes | No — no custom LaTeX templates |
| 1.8.x | No breaking changes documented | N/A |
| 1.9.x | highlight-style deprecated → syntax-highlighting | No — not used |
| 1.9.x | cookie-consent default changed to express | No — not used |
| 1.9.x | RevealJS scroll-view template changes | No — no presentations |
| 1.9.x | ^[ footnote syntax conflict with superscript spans | No — not used in any .qmd |
| 1.9.x | Stricter output-dir validation (must be subdirectory) | OK — ours is _build, a subdirectory |

## Example of Compare Versions Report

[compare-versions-2026-04-14T15-08-16.md](https://github.com/user-attachments/files/26722224/compare-versions-2026-04-14T15-08-16.md)

Because of localhost loading, browser/network issues, etc., a perfect comparison is not possible, but it does float up interesting issues.
I looked through what it generated, and nothing is actually a relic of the Quarto 1.9 upgrade.

## Additional Notes

I split the "Audit Site" job into another branch to do a CI/CD run, because that should be able to run on a schedule to do some better detection of issues.


cpsievert:
Thanks! I'm slightly concerned about the potential for breaking changes with the Quarto upgrade. To help ease those concerns, maybe we could use something like https://github.com/vercel-labs/agent-browser to try and catch any regressions? Could also feed the Quarto release notes to Claude Code and have it investigate whether we're using any features that have been removed?

CLAUDE.md
cpsievert, line 1: I think it would make some sense to include a section on the tests available in `CLAUDE.md`? I think it could also make some sense to provide some guidance on when to run the tests (i.e., always run comparison when the Quarto version changes)?

tests/compare-versions.ts
```diff
@@ -0,0 +1,331 @@
...
+    }
+  }
+
+  if (visualChanges.length) {
+    console.log(`\n## Visual Changes (${visualChanges.length})\n`);
+    for (const d of visualChanges) {
+      console.log(`  ${d.path}`);
+      console.log(`  ${d.summary}\n`);
+    }
+  }
```
cpsievert: Anything interesting or notable that Claude was able to find?

elnelson575: So I ended up just running the visual on audit-site.ts and it did find a few things-- actually entirely stuff unrelated to the upgrade, but that got missed, so I've been making some issues to document those.
I've been looking at the compare-version now and seeing how vercel agent-browser might work there to reduce the reliance on Claude vision. Do we have an opinion on how many tokens we're interested in spending on Claude vision? I can't see Bedrock usage, but doing the compare with vision would double the tokens, I think (send the old and the new screenshot) and I know image processing is pricey.

CLAUDE.md
elnelson575: Yep! Finishing refining out the scripts and then I'll add that in! 👍🏻

tests/compare-versions.ts
cpsievert: I think as long as we're sending a relatively low resolution to a relatively cheap model (haiku or sonnet) then I don't think it's worth worrying about too much.

elnelson575: If Claude is to be believed, based on resolution and number of pages it estimated ~$0.60 for haiku 4.5 and $2.25 for Sonnet 4.6, so I think we can spare less than $3 for Sonnet for a job we would run only infrequently 😆 
I'll see how it does and if it's unimpressive, I might add an option to increase the resolution for super occasional regression tests.

cpsievert:
> but it does float up interesting issues.

Let's make sure to file issues on what it has surfaced to make sure we don't forget.

cpsievert (approved):

Issue #409, opened 2026-07-24
Document remaining shiny.ui exports missing a component/layout page


schloerke:
Follow-up to #408, which added `components/test_ui_api_has_page.py` — a test that requires every public `shiny.ui` / `shiny.express.ui` export to be documented by a page's `relevant-functions` block or be explicitly opted out. The exports below are opted out today because they have no page yet. Each group is a named set in that test file; when a function gets documented, delete it from the set and the test starts enforcing it.

`page_*` functions are intentionally excluded from this issue for now.

## Components needing a page (`_TODO_NEEDS_COMPONENT_PAGE`)

- [x] `download_button` — #416
- [x] `download_link` — #416
- [x] `popover` — #412
- [x] `chat_ui` — #418
- [x] `output_code` — #418
- [x] `output_table` — #418
- [x] `output_markdown_stream` — #418
- [x] `toast` — #413
- [x] `show_toast` — #413
- [x] `hide_toast` — #413
- [x] `toast_header` — #413
- [x] `input_task_button` — #415
- [x] `bind_task_button` — #415
- [x] `input_code_editor` — #417
- [x] `input_submit_textarea` — #417
- [x] `input_bookmark_button` — #415
- [ ] `offcanvas` — #422
- [ ] `show_offcanvas` — #422
- [ ] `hide_offcanvas` — #422
- [ ] `toggle_offcanvas` — #422

Offcanvas arrived later than the rest of this list: it is new in shiny 1.7.0
(posit-dev/py-shiny#2279) and was opted out into `_TODO_NEEDS_COMPONENT_PAGE` by #422 to
unblock CI on the v1.7.0 submodule bump (#421). The `Offcanvas` class is opted out
separately in `_CLASSES` and should be documented alongside these four on the same page.

## Nav / panel functions needing a layout-page entry (`_TODO_NEEDS_LAYOUT_PAGE`, excluding `page_*`)

- [x] `nav_spacer` — #414
- [x] `navbar_options` — #414
- [x] `navset_bar` — #414
- [x] `navset_card_underline` — #414
- [x] `navset_hidden` — #414
- [x] `navset_underline` — #414
- [x] `panel_conditional` — #414

## Mutators whose base component has no page yet (`_MUTATORS`)

Add each to its base component's `relevant-functions` block when that page is written:

- [x] `update_code_editor` (base: `input_code_editor`) — #417
- [x] `update_popover` (base: `popover`) — #412
[...]

PR #386, opened 2026-07-22
docs: Add download-button component docs (render.download_button)

schloerke:
## Summary

Supersedes #351, rebased onto the 1.7.0 release staging branch (`rc-shiny-v1.7.0`) instead of `main`.

- Adds the download-button component documentation page (`components/inputs/download-button/index.qmd`) and its example apps (core, express, preview, detail-preview).
- Migrates all examples from the deprecated `@render.download` to `@render.download_button` (py-shiny#2364). `ui.download_button(...)` UI calls are unchanged.
- Updates the API listing on the component page to reference `@render.download_button` / `@render.download_link` (dropping the old `@render.download` entry), and adds a "See also" cross-reference to the download-link component page.
- Adds sidebar nav entries in `_quarto.yml` for **both** `components/inputs/download-button/index.qmd` and `components/inputs/download-link/index.qmd`. The download-link page content itself is added by a companion PR from a sibling agent; this PR owns the shared `_quarto.yml` nav entries for both pages since only one PR can touch that file without conflicts.

## Test plan
- [ ] `_quarto.yml` parses as valid YAML (verified locally with Ruby's YAML parser; no Python yaml module available in this sandbox)
- [ ] `grep -rn "render\.download\b" components/inputs/download-button` returns nothing
- [ ] Once the companion download-link PR merges, confirm the site renders both new sidebar entries and both pages build correctly

## TODO before merge

- [ ] Add `thumbnail.png` screenshot for the download-button kitchen-sink example (`components/inputs/download-button/thumbnail.png`) — could not be generated headlessly in this sandbox; the `image: thumbnail.png` line in `index.qmd`'s `kitchen-sink` listing is marked with an inline `# TODO: release-docs` comment.
- [ ] **Do not merge until the site's `py-shiny` submodule is bumped to include `render.download_button`/`render.download_link`** (py-shiny#2364 → 1.7.0 release). Until then quartodoc won't generate the new API-reference pages and the API links here are dead.
- [ ] **Interactive shinylive examples need a `shinylive` version bundling shiny ≥ 1.7.0** — `requirements.txt` pins `shinylive==0.8.9` (pre-1.7.0 shiny, no `@render.download_button`); bump when available.
- [ ] Merge together with the companion download-link PR #385 — this PR adds the `_quarto.yml` nav entry for the download-link page that #385 provides, so merging alone would reference a not-yet-existing page.

PR #346, opened 2026-03-03
Feat/document components

elnelson575:
Document components that are currently in the codebase for py-shiny but do not have dedicated documentation on the components page.

PR #416, opened 2026-07-25
docs(components): add Download Button and Download Link component pages

karangattu:
Part of #409.

Adds component pages for the two download controls under **Components > Inputs**:

- `components/inputs/download-button/` — documents `ui.download_button`
- `components/inputs/download-link/` — documents `ui.download_link`

Both pages document the **current stable API on main (shiny 1.6.3)**: `ui.download_button()` / `ui.download_link()` paired with the `@render.download` decorator. The example apps generate a small CSV in memory via `yield`, so they run under shinylive with no filesystem or network assumptions. The download-link Express example uses the `ui.hold()` + `shiny.ui.download_link()` pattern, since 1.6.3's `@render.download` auto-places a *button* in Express mode.

Relationship to other PRs:

- #385 / #386 (schloerke) document the new v1.7.0 `render.download_button` / `render.download_link` API and target the `rc-shiny-v1.7.0` branch. That API does not exist in 1.6.3, so these pages are written against the current API; #385/#386 can rebase onto / update these pages for the v1.7.0 API when the RC lands. Structure and prose here borrow from those PRs.
- #346 (elnelson575) included an earlier download-button page attempt as part of an umbrella PR.

Other changes:

- Registers both pages alphabetically in the `_quarto.yml` inputs sidebar.
- Removes `download_button` / `download_link` from `_TODO_NEEDS_COMPONENT_PAGE` in `components/test_ui_api_has_page.py`.
- Shinylive links and `relevant-functions` fields are generated (`make components-shinylive-links`, `make components-relevant-functions`); `make components-static-previews` renders both preview cards.

Testing:

- New Playwright interaction tests (`test_download_button.py`, `test_download_link.py`) drive the Core AND Express apps with `controller.DownloadButton` / `controller.DownloadLink`, click the control, and assert the downloaded file's suggested filename and exact CSV content via `page.expect_download()` — 4 passed.
- `uv run pytest components/test_examples_smoke.py -k "download-button or download-link" -n0` — 8 passed (all 8 new example apps).
- `uv run pytest components/test_ui_api_has_page.py -n0` — 4 passed.
- Both pages render cleanly with `quarto render`.

### DIFF
--- a/_quarto.yml
+++ b/_quarto.yml
@@ -148,6 +148,8 @@ website:
             - components/inputs/dark-mode/index.qmd
             - components/inputs/date-range-selector/index.qmd
             - components/inputs/date-selector/index.qmd
+            - components/inputs/download-button/index.qmd
+            - components/inputs/download-link/index.qmd
             - components/inputs/file/index.qmd
             - components/inputs/numeric-input/index.qmd
             - components/inputs/password-field/index.qmd
--- a/components/inputs/download-button/app-core.py
+++ b/components/inputs/download-button/app-core.py
@@ -1,0 +1,18 @@
+from datetime import date
+
+from shiny import App, render, ui
+
+app_ui = ui.page_fluid(
+    ui.download_button("download_data", "Download CSV"),  # <<
+)
+
+
+def server(input, output, session):
+    @render.download(filename=lambda: f"data-{date.today().isoformat()}.csv")
+    def download_data():
+        yield "name,value\n"
+        yield "Alice,100\n"
+        yield "Bob,200\n"
+
+
+app = App(app_ui, server)
--- a/components/inputs/download-button/app-detail-preview.py
+++ b/components/inputs/download-button/app-detail-preview.py
@@ -1,0 +1,20 @@
+from datetime import date
+
+from shiny import App, render, ui
+
+app_ui = ui.page_fluid(
+    ui.h5("Download example data"),
+    ui.download_button("download_data", "Download CSV"),
+    ui.p("Click the button to download a CSV file generated in memory."),
+)
+
+
+def server(input, output, session):
+    @render.download(filename=lambda: f"data-{date.today().isoformat()}.csv")
+    def download_data():
+        yield "name,value\n"
+        yield "Alice,100\n"
+        yield "Bob,200\n"
+
+
+app = App(app_ui, server)
--- a/components/inputs/download-button/app-express.py
+++ b/components/inputs/download-button/app-express.py
@@ -1,0 +1,13 @@
+from datetime import date
+
+from shiny.express import render
+
+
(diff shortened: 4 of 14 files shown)


schloerke:
Closing in favour of #385 (download-link) and #386 (download-button), which document the current 1.7.0 API. Thanks @karangattu — this wasn't wasted, and I've carried the good parts across.

## Why not this PR

All 8 app files use `render.download`, which is explicitly deprecated in py-shiny (`shiny/render/_render.py`): `render.download is deprecated. Use render.download_button (or render.download_link) instead.` That isn't a cosmetic issue — the smoke sweep treats deprecation warnings on stderr as failures, so `smoke (2)(3)(4)(5)` and `test-relevant-functions` all went red once the branch was updated from `main`. #385/#386 were already open against `main` using `render.download_button` / `render.download_link`, and only one set can land in these two directories.

## What I lifted from here

**Your tests, which were the best thing in this PR.** #385/#386 had *no* tests at all. The pattern you wrote — click, assert `suggested_filename`, then compare exact downloaded bytes — is exactly right, including the detail that `download.path()` is unavailable against CI's remote browser so you have to `save_as()` a local copy. I kept that comment. I extended them to cover both kitchen sinks (`width=`, `icon=`, `class_=`).

I also took your `test_ui_api_has_page.py` change dropping `download_button` / `download_link` from the opt-out set — both #385 and #386 had left theirs in place.

Landed as `910b6cb` (#385) and `7c435d9` (#386).

## One thing you were right about that I got wrong

My review said the `ui.hold()` workaround in your Express example was "obsolete and actively misleading." **That was too broad.** It's obsolete for the *basic* case, where `render.download_link()` auto-places the link. But `render.download_link()` / `render.download_button()` accept only `filename`, `media_type`, `encoding`, `label` and `width` — there is no `icon` parameter, and no way to pass `class_`/`style`. For those, explicit placement plus `ui.hold()` to suppress the auto-placed duplicate is the *only* way, and both kitchen sinks in #385/#386 now use precisely that pattern, with a mutation-verified test that fails if the `ui.hold()` is removed.

So the workaround you documented is real and still needed; I was wrong to call it obsolete outright. It just shouldn't be the headline example.

## Also worth knowing

`shiny.express.ui` has **no** download functions at all — not `download_button`, not `download_link`. Both kitchen sinks in #385/#386 were calling `ui.download_button(...)` after `from shiny.express import ui` and **failed to start** (`module 'shiny.express.ui' has no attribute 'download_button'`). They have to be imported from `shiny.ui`. Your PR sidestepped this by not having Express kitchen sinks, so this is a bug you didn't hit.

The remaining gaps I noted in review (no variations, `icon=`/`width=` unexercised) are now covered by the kitchen sinks and their tests.

PR #436, opened 2026-08-06
docs: document resolve_id() and current_namespace() on the modules page

jat255:
Two documentation requests from posit-dev/py-shiny#2019 and posit-dev/py-shiny#2127 that touch the same page (`docs/modules.qmd`), so they're bundled here as one PR.

## posit-dev/py-shiny#2019 — `resolve_id()` for ids in custom HTML

The reporter expected `@module.ui` to namespace an `id` passed to `ui.div()`. It wasn't obvious that this is not how that decorator works, so it's worth updating the module docs to reflect how this works in reality.

New section **"Namespacing ids in custom HTML"** explains that `@module.ui` does not rewrite the UI it receives. It only sets the current namespace, and it's each `ui.input_*()` / `ui.output_*()` function that looks the namespace up and prefixes its own id. That's why a hand-written `id=` passes through untouched, and why `resolve_id()` works around the issue. Includes the issue's before/after snippets and an Express/Core app where the resolved id genuinely matters (each module instance inserts into its own `ui.insert_ui()` target).

## posit-dev/py-shiny#2127 — Making use of `current_namespace()` more explicit

New section **"Getting the current module id"** documents `current_namespace()`, including that the value is the fully resolved namespace (`"out-inner"` for a module nested inside `outer_ui("out")`), that it's a `str` subclass, that it works in server functions as well as UI functions, and that it's empty outside a module. The Express/Core example is the reporter's case: label and output vary by module id, with no extra parameter.

## Verification

- `shiny.module.resolve_id` and `shiny.module.current_namespace` both confirmed present and in `shiny.module.__all__` on released shiny 1.7.0.
- Every snippet was executed to verify the outputs against current shiny. The two `print(modui("modid"))` examples produce exactly the `<div id="divid">` / `<div id="modid-divid">` output shown.
- All four example apps run under `shiny run`, and the interactive behavior was checked with Playwright: notes insert into the correct per-instance container, and the `row_3`/`row_5` labels and per-row outputs are correct.
- `quarto render docs/modules.qmd` succeeds; both new sections and all shinylive blocks render.

## Note for reviewers

- One cross-link added, from the new `resolve_id()` section to [Custom JavaScript component](https://shiny.posit.co/py/docs/custom-component-one-off.html), which already uses `resolve_id()` for exactly this reason. No other page was touched.
- `current_namespace` is exported in `shiny.module.__all__` (since 2022) but has no docstring and isn't in the API reference (`_quartodoc-core.yml` lists only `module.ui` and `module.server`). Documenting it here as public seems right given the maintainer recommended it on the issue, but adding a docstring + reference entry in py-shiny would be a reasonable follow-up.
- The rendered documentation has little red squiggles for `resolve_id` and `current_namespace` imports, which is this issue that needs a separate fix (it still works, it's just an editor nit): https://github.com/posit-dev/py-shiny/issues/2425

## Screenshots

image

image

image

image

### DIFF
--- a/docs/modules.qmd
+++ b/docs/modules.qmd
@@ -423,16 +423,18 @@ Since modules allow you to tie UI and Server code together in the same namespace
 Anything that you can do in a Shiny app can also be done inside of a module, and modules can themselves call other modules.
 This allows you to break your app up into building blocks of various sizes, compose those blocks to build different applications, and share them with others.
 
-## Namespacing ids in custom HTML
+## Namespacing IDs in custom HTML
 
-Shiny's own UI functions take care of namespacing for you.
-Call `ui.input_text("text_in", "Enter text")` inside a module whose id is `row_1`, and the id that reaches the browser is `row_1-text_in`.
+Shiny's UI functions handle namespacing for you. If you call `ui.input_text("text_in", "Enter text")`
+inside a module with the id `row_1`, the ID that reaches the browser is `row_1-text_in`.
 
-It's worth being precise about *where* that happens, because it explains a behavior that otherwise looks like a bug.
-The `@module.ui` decorator does not rewrite the UI that your function returns.
-All it does is set the *current namespace* for the duration of the call; it is each individual `ui.input_*()` and `ui.output_*()` function that looks up that namespace and prefixes its own id with it.
+It's worth going into some detail about how that happens, because it explains a behavior which might look like
+a bug at first glance. The `@module.ui` decorator does not modify the UI that your function returns.
+Instead, it sets the current namespace while the function runs, and then each `ui.input_*()` and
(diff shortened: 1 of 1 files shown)


cpsievert:
Nice, thanks! Two questions

1. Were you able to preview this locally? 
2. Was there a reason why you needed to submit this from your own personal fork? 

I think I can add you as a collaborator, but I'm not sure if that's needed.  Also, if you submit the PR from the main repo, then I think you'll get a preview button.

jat255:
Thanks for the review on this. Responses:

1. Yes, I previewed it locally (that's the screenshots in the description)
2. I had set up my remote on this repo when I didn't have access, so this went to my fork by default. This PR is superseded by #438, which has the same branch and commits.

PR #437, opened 2026-08-06
docs(components): improve gallery previews and component page UX

karangattu:
## Summary

Improves the component gallery and component detail-page experience.

### Component gallery

- Replace loading Shinylive embeds with lightweight static thumbnails.
- Animate each component preview on hover or keyboard focus.
- Use the GIF’s exact first frame as its static thumbnail.
- Add preview assets for all 53 components.
- Display three cards per row on desktop and one per row on mobile.
- Link the entire preview card to its component page.
- Respect touch devices and reduced-motion preferences.
- Add asset-versioned URLs so updated thumbnails and GIFs do not remain cached.
- Reframe previews for consistent sizing and readable interactions.
- Fix the Select (Single) preview so it demonstrates Choice 1A → 1B → 1C without an unresolved Shiny output or spinner artifacts.

### Component pages and navigation

- Disable copy-to-clipboard buttons on component pages while leaving the rest of the site unchanged.
- Preserve the “Edit in Shinylive” links when copy buttons are absent.
- Order input components consistently with the sidebar:
  Select (Single) → Select (Multiple) → Selectize (Single) → Selectize (Multiple).
- Add a regression test requiring every input component page to appear in the explicit gallery listing.


schloerke:
## Thumbnails

gallery-thumbnails

## Gifs

https://github.com/user-attachments/assets/8f6f2e84-588c-4228-a498-f43f8b96ae92

PR #440, opened 2026-08-11
perf(components): defer offscreen previews

karangattu:
## Summary

This change defers component previews that are not near the viewport. Each preview starts when the user scrolls near it.

This reduces the work during the first page visit.

## Changes

- Static preview iframes start near the viewport.
- Shinylive apps start near the viewport.
- Preview cards keep a stable height while they load.
- The loading effect respects reduced-motion preferences.
- Preview iframes have accessible titles.
- Browser tests cover preview startup and scroll behavior.

## Results

Local browser measurements show these results:

- The initial preview count decreased from 43 to 9.
- The initial resource request count decreased from 100 to 61.
- The remaining previews start as the user scrolls.

## Tests

- All 131 relevant tests passed.
- `quarto render components/index.qmd` completed without errors.
- Browser tests showed correct behavior before and after scrolling.

### DIFF
--- a/components/_partials/components-list.css
+++ b/components/_partials/components-list.css
@@ -82,6 +82,40 @@ div.shinylive-wrapper {
   justify-content: center;
   align-content: center;
   flex-wrap: wrap;
+  content-visibility: auto;
+  contain-intrinsic-size: 150px;
+}
+
+.component-list-card:has(iframe[data-src]),
+.component-list-card:has(.deferred-shinylive-python) {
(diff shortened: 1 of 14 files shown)


schloerke:
Closing: #437 replaced the gallery's live preview iframes with static thumbnails + hover GIFs, and #470 removed the static preview HTML and `static: true` flags, so the deferred-iframe loading here no longer has anything to defer.

Two parts are still relevant on their own and are being carried forward as separate stacked PRs (with you as co-author, @karangattu — thank you!):
- the flaky `plot-plotly` interaction test fix
- the site-wide Shinylive service-worker reload-loop fix in `include-in-header.html` + its test

PR #443, opened 2026-08-12
fix: turn off "edit this page" on generated api/ pages

### LINKED ISSUE #187: Turn off "edit this page" under `api/`

### LINKED ISSUE #280: "edit this page" links break on the reference pages
Seems like the "edit this page" link is broken for the reference pages. Examples:

- express: <https://shiny.posit.co/py/api/express/express.ui.input_action_link.html>
- core: <https://shiny.posit.co/py/api/core/ui.layout_column_wrap.html>

My suspicion is the location of the `qmd` file in the repo isn't the same folder from a redirect.

schloerke:
Fixes #187
Fixes #280

## Problem

Pages under `api/` are generated at build time by quartodoc and are **not** committed to this repo (`api/.gitignore` ignores `*.qmd`). The page toolbox still rendered an "Edit this page" link, e.g.

    https://github.com/posit-dev/py-shiny-site/edit/main/api/core/ui.layout_column_wrap.qmd

which 404s on GitHub.

## Fix

Add `api/_metadata.yml` with `repo-actions: false`, which Quarto applies to every document in `api/` and its subdirectories.

Note: Quarto only honors `true`/`false` for a **document-level** `repo-actions` — a list is ignored, since the action list always comes from `website.repo-actions` in `_quarto.yml` (see `handleRepoLinks` in the Quarto source). So this turns off the whole toolbox under `api/` ("Report an issue" included) rather than just the `edit` action. The three hand-written `api/*/index.qmd` pages lose their toolbox too; their bodies are almost entirely a generated `{{< include _api_index.qmd >}}`, so an edit link there was not much more useful.

## Verification

Rendered with the pinned Quarto 1.9.36 and grepped the output:

| page | toolbox |
| --- | --- |
| `api/core/ui.layout_column_wrap.html` | none |
| `api/express/express.ui.input_action_link.html` | none |
| `api/core/index.html` | none |
| `docs/modules.html` (control) | "Edit this page" + "Report an issue" still present |

`make quartodoc` copies generated files in with `rsync -ac` (no `--delete`), so the new tracked `_metadata.yml` survives regeneration.

### DIFF
--- a/api/_metadata.yml
+++ b/api/_metadata.yml
@@ -1,0 +1,62 @@
+# Turn off the repo-action toolbox ("Edit this page" / "Report an issue") for
+# every page under api/ and its subdirectories.
+#
+# WHY
+# ---
+# Pages under api/ are generated at build time by quartodoc and are not
+# committed to this repository (api/.gitignore ignores *.qmd), so the "Edit
+# this page" action linked to a file that does not exist on GitHub and 404'd:
+#
+#   https://github.com/posit-dev/py-shiny-site/edit/main/api/core/ui.layout_columns.qmd
+#
+# See https://github.com/posit-dev/py-shiny-site/issues/187
+# and https://github.com/posit-dev/py-shiny-site/issues/280
+#
+# WHY `false` AND NOT `[issue]`
+# -----------------------------
+# Quarto only honors `true`/`false` for a *document-level* `repo-actions`. A
+# list is silently ignored here: the set of actions always resolves from
+# `website.repo-actions` in _quarto.yml (currently `[issue, edit]`). Quarto's
+# `handleRepoLinks` tests the document value for `=== false` (suppress all) and
+# `=== true` (force), then reads the list from the website config. The docs
+# likewise document only `repo-actions: false` as the per-page knob:
+#
+#   https://quarto.org/docs/websites/website-navigation.html#github-links
+#
+# `website.issue-url` does not help either -- it is consulted *after* the
+# `=== false` early return, so it cannot re-add the issue link on a page whose
+# toolbox has been suppressed.
+#
+# WHAT THIS COSTS
+# ---------------
+# Diffing an old vs. new render of api/core/ui.layout_columns.qmd, exactly two
+# blocks disappear -- two copies of the same link list:
+#
+#   1. <div class="toc-actions"> at the bottom of the right-hand "On this page"
+#      TOC nav.
+#   2. Its small-screen duplicate, <div class="toc-actions d-sm-block d-md-none">
+#      inside .nav-footer-center (that footer cell is now just &nbsp;).
+#
+# Each held two links:
+#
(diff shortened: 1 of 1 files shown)


schloerke:
## What is lost

Diffed an old vs. new render of `api/core/ui.layout_columns.qmd`. Exactly two blocks disappear, both copies of the same link list:

1. `<div class="toc-actions">` at the bottom of the right-hand "On this page" TOC nav.
2. The small-screen duplicate, `<div class="toc-actions d-sm-block d-md-none">` inside `.nav-footer-center` (that footer cell is now just `&nbsp;`).

Each held two links:

| link | status |
| --- | --- |
| **Edit this page** → `/edit/main/api/core/ui.layout_columns.qmd` | **already broken** (404) — the bug being fixed |
| **Report an issue** → `/issues/new` | **worked; this is the real loss** |

`source` ("View source") was never enabled — `_quarto.yml` sets `repo-actions: [issue, edit]` — so nothing is lost there. Nothing else on the page changes: TOC entries, sidebar, breadcrumb, footer links, and page navigation are all untouched. I also grepped for `data-quarto-source-url="repo"` (the other thing `handleRepoLinks` early-returns past) — no occurrences in this site, so that branch is a no-op here.

**Net: "Report an issue" is gone from the ~200 generated `api/` pages plus the 3 hand-written `api/*/index.qmd` pages.**

There is no narrower knob. Per [the Quarto docs](https://quarto.org/docs/websites/website-navigation.html#github-links), page-level control is documented only as `repo-actions: false`; the action *list* is a website-level option. That matches `handleRepoLinks` in the Quarto source, which checks the document value for `=== false` / `=== true` and otherwise reads the list from `website.repo-actions`.

If losing "Report an issue" here is not acceptable, the option is to re-add it independently of the repo-actions machinery (e.g. an `include-after-body` or a footer tweak scoped to `api/`). Happy to do that — say the word.

schloerke:
Note on precedent (I missed this earlier): `repo-actions: false` in a directory `_metadata.yml` is **already the established convention in this repo**, not a new mechanism.

| `_metadata.yml` | `repo-actions` | toolbox on the live site |
| --- | --- | --- |
| `components/` | `false` | no |
| `layouts/` | `false` | no |
| `gallery/` | `false` | no |
| `templates/` | `false` | no |
| `docs/` | *(unset)* | **yes** — inherits `[issue, edit]` |
| `get-started/` | *(unset)* | **yes** — inherits `[issue, edit]` |
| `api/` | `false` *(this PR)* | no |

Verified on production: `components/inputs/action-button/`, `layouts/` → 0 toolbox links; `docs/reactive-foundations.html`, `docs/modules.html` → both links present.

This also softens the "we lose Report an issue" caveat I raised above: four directories already make exactly that trade, so `api/` joining them is consistent with how the site is already configured rather than a new precedent.

